PR #6, opened 2022-01-14
Change besseli0 and besseli1 routines to rational approximation

heltonmc:
Per a comment from @oscardssmith. Using the chebyshev basis is significantly slower than monomial basis.

This PR implements `besseli0` and `besseli1` using monomial basis following the derivations and coefficients listed [here](https://www.advanpix.com/2015/11/11/rational-approximations-for-the-modified-bessel-function-of-the-first-kind-i0-computations-double-precision/) and [here](https://www.advanpix.com/2015/11/12/rational-approximations-for-the-modified-bessel-function-of-the-first-kind-i1-for-computations-with-double-precision/).

This results in a significant speed up (2.5x than master) without loss of accuracy.

```julia
### old version on master branch
julia> @benchmark besseli0(x) setup=(x = 100*rand())
BenchmarkTools.Trial: 10000 samples with 992 evaluations.
 Range (min … max):  36.447 ns … 50.752 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     36.858 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   37.172 ns ±  0.861 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▅▆██       ▁▄▄                 ▁▁▁                  ▅▄ ▂
  ▃▄▅▁▁████▇▁▁▃▁▃▄███▄▃▁▁▁▁▁▁▁▁▃▄▄▅▅▆█████▇▆▆▆▅▆▆█▅▅▃▄▃▅▄▃▅██ █
  36.4 ns      Histogram: log(frequency) by time      39.5 ns <

### this PR
julia> @benchmark besseli0(x) setup=(x = 100*rand())
 Range (min … max):   5.842 ns … 27.094 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     13.741 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.153 ns ±  2.163 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                   █           
  ▃▁▂▁▁▁▁▁▁▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▃▂▂▂▁▁▁▂▂ ▂
  5.84 ns         Histogram: frequency by time        15.3 ns <

```

and comparing again to `SpecialFunctions.jl` it's about 18x faster.
```julia
using SpecialFunctions
julia> @benchmark besseli(1, x) setup=(x = 100*rand())
BenchmarkTools.Trial: 10000 samples with 937 evaluations.
 Range (min … max):   94.832 ns … 886.260 ns  ┊ GC (min … max): 0.00% … 37.42%
 Time  (median):     249.701 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   306.212 ns ± 130.085 ns  ┊ GC (mean ± σ):  0.03% ±  0.87%

   ▂▃▂         ██▇▆▆▆▅▅▄▄▃▂▃▃▂▂▁▁▁▁▂▁▁▁▁▂▂▁▁▂▁▂▂▁▂▂▂▂▂▃▁▃▂▂▃▂▂▂ ▃
  ████▁▁▁▁▁▁▁▁▁████████████████████████████████████████████████ █
  94.8 ns       Histogram: log(frequency) by time        670 ns <

```


oscardssmith:
Nice! How many ULPs did you get before vs after?

heltonmc:
So tested against `ArbNumerics.jl` using 50 digits of precision for 10000 evenly spaced points between 0 and 100 the mean and max relative errors are as follows for `besseli0(x)`:

Old versions
Mean relative error: 1.00800350195715708e-16
Max relative error: 3.75163868123101539944e-16

New version
Mean relative error: 1.184275524171108551677614992624e-16
Max relative error: 4.036293772497829275701727546031e-16

So a very slight increase in error.
Is there a built in function to calculate ULP from the residuals?

src/besseli.jl
```diff
@@ -1,46 +1,105 @@
...
-    if x <= 8
-        y = muladd(x, T(.5), T(-2))
-        return exp(x) * chbevl(y, A_i0(T))
+    T == Float32 ? branch = 50 : branch = 500
+    x = abs(x)
+    if x < 7.75
+        a = x * x / 4
+        return muladd(a, evalpoly(a, P1_i0(T)), 1)
+    elseif x < branch
+        return exp(x) * evalpoly(inv(x), P2_i0(T)) / sqrt(x)
```
oscardssmith, line 22: I don't think this branch is worthwhile the branch for large `x` handles medium `x` well (and cheaply). Also, recalculating these coefficients so you are using `exp2` rather than `exp` would save you a small amount of time and precision.

heltonmc: The issue with `besseli` and `besselk` is their diverging properties. This small branch is to prevent overflow in exp(x) before there is overflow in `besseli`. 
So for example.
```julia
### using SpecialFunctions
julia> besseli(0, 710.0)
ERROR: AmosException with id 2: overflow.
Stacktrace:
 [1] _besseli(nu::Float64, z::ComplexF64, kode::Int32)
......

##### using this code
julia> besseli0(710.0)
3.3453345586196564e306
```
So we can see that the `besseli` implementation in SpecialFunctions.jl does not handle this zone properly and overflows before `besseli` actually does. So it is a rather small branch but it is able to provide more accurate results when you reach closer to floatmax.

Do you think this is worth it?

oscardssmith: what I'm saying is to use the branch that you use for large numbers for all numbers, since the `exp(x/2)` will work for the whole range.

heltonmc: Ahh I see! Ok so if we change it by removing the `P3_i0` branch
```julia
function besseli0(x::T) where T <: Union{Float32, Float64}
    x = abs(x)
    if x < 7.75
        a = x * x / 4
        return muladd(a, evalpoly(a, P1_i0(T)), 1)
    else
        a = exp(x / 2)
        s = a * evalpoly(inv(x), P2_i0(T)) / sqrt(x)
        return a * s
    end
end
```
This will pass all tests.
The trade off becomes `P2_i0` is length 22 whereas `P3_i0` is length 2 so large arguments will take a little more time to compute. 
Ok with that? I can do some benchmarks

src/constants.jl
```diff
@@ -104,51 +104,67 @@ const QQ_j0(::Type{Float64}) = (
...
-    7.51729631084210481353E-18, 4.41434832307170791151E-18, -4.65030536848935832153E-17,
-    -3.20952592199342395980E-17, 2.96262899764595013876E-16, 3.30820231092092828324E-16,
-    -1.88035477551078244854E-15, -3.81440307243700780478E-15, 1.04202769841288027642E-14,
-    4.27244001671195135429E-14, -2.10154184277266431302E-14, -4.08355111109219731823E-13,
-    -7.19855177624590851209E-13, 2.03562854414708950722E-12, 1.41258074366137813316E-11,
-    3.25260358301548823856E-11, -1.89749581235054123450E-11, -5.58974346219658380687E-10,
-    -3.83538038596423702205E-9, -2.63146884688951950684E-8, -2.51223623787020892529E-7,
-    -3.88256480887769039346E-6, -1.10588938762623716291E-4, -9.76109749136146840777E-3,
-    7.78576235018280120474E-1
+const P1_i0(::Type{Float32}) = (
```
oscardssmith, line 107: I would also recommend renaming these to `besseli0_small_coefs` or something similar.

src/besseli.jl
heltonmc: Some benchmarks
```julia
# removing the third branch
julia> @benchmark besseli0(x) setup=(x=rand(500:0.0001:700))
 Range (min … max):  22.925 ns … 63.379 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     23.050 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   23.092 ns ±  0.629 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

               █     █      ▁                                  
  ▂▁▁▁▁▁▆▁▁▁▁▁▁█▁▁▁▁▁█▁▁▁▁▁▁█▁▁▁▁▁█▁▁▁▁▁▁▅▁▁▁▁▁▃▁▁▁▁▁▁▂▁▁▁▁▁▂ ▂
  22.9 ns         Histogram: frequency by time        23.3 ns <

# keeping the third branch
julia> @benchmark besseli0(x) setup=(x=rand(500:0.0001:700))
 Range (min … max):  12.053 ns … 25.985 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     12.137 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   12.186 ns ±  0.474 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

           █    ▁                                              
  ▂▁▁▁▄▃▁▁▁█▁▁▁▁█▁▁▁▁▄▁▁▁▁▃▁▁▁▁▂▁▁▁▁▂▁▁▁▁▂▁▁▁▁▁▁▁▁▁▂▁▁▁▁▂▁▁▁▂ ▂
  12.1 ns         Histogram: frequency by time        12.6 ns <

```

I also tested for smaller arguments and they are roughly similar.
```julia
julia> @benchmark besseli02(x) setup=(x=rand(0.00001:0.0001:100))
 Range (min … max):  10.594 ns … 48.674 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     22.690 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   21.880 ns ±  3.261 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                     █▁        
  ▄▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁██▂▂▂▂▂▂ ▂
  10.6 ns         Histogram: frequency by time        24.4 ns <

julia> @benchmark besseli0(x) setup=(x=rand(0.00001:0.0001:100))
 Range (min … max):   9.635 ns … 40.666 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     22.731 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   21.822 ns ±  3.526 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄                                                  ▃█  ▂    ▁
  █▇▄▄▁▃▁▁▁▁▁▁▁▁▁▃▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁███▅█▇▇▇ █
  9.63 ns      Histogram: log(frequency) by time      24.4 ns <

```

oscardssmith: I think that's probably a good compromise (but I'm not 100% sure and it probably depends on how frequently you call  the function with big arguments.

oscardssmith: To see the cost of the branch, you're going to want to test a mix of above and below (unsorted) so the branch predictor can't save you

oscardssmith: but your current version might just be the way to go.

heltonmc: One thing I don't have experience in is optimizing this for SIMD better. It would be nice to have these play well with LoopVectorization.. Unfortunately, I've frequently had to compute bessel functions for large arguments. They are often used in integral transforms requiring a large number of evaluations to reach good convergence (but this could just be a very special case where `besseli(x)` is needed in an ordered array from 0 -> inf).

I'll do some more benchmarks and leave this open for a bit! If you have any resources for providing better SIMD-able functions let me know. It could be its well worth dropping this branch if optimizations can be made elsewhere.

oscardssmith: tldr is that for a simd implementation, you end up calculating both sides of the branch, so it will definitely be bad.

heltonmc: I'll propose removing this branch for `besseli0` and `besseli1` and keeping it for the scaled versions. In general, if you are anticipating large arguments you should be designing your code to use the scaled versions. So if you are using `besseli0x` you are much more likely to call it for very large arguments. Maybe next I can work on the docs 😅

src/besseli.jl
```diff
@@ -1,46 +1,99 @@
...
+    if x < 7.75
+        a = x * x / 4
+        return muladd(a, evalpoly(a, besseli0_small_coefs(T)), 1) * exp(-x)
+    elseif x < branch
+        return evalpoly(inv(x), besseli0_med_coefs(T)) / sqrt(x)
+    else
+        return evalpoly(inv(x), besseli0_large_coefs(T)) / sqrt(x)
+    end
+end
+function besseli1(x::Float32)
```
oscardssmith, line 37: This `Float32` implimentation is exactly the same as the `Float64` version, should they be merged?

src/besseli.jl
```diff
@@ -1,46 +1,99 @@
...
+        s = a * evalpoly(inv(z), besseli1_med_coefs(T)) / sqrt(z)
+        z =  a * s
+    end
+    if x < zero(x)
+        z = -z
+    end
+    return z
 end
-function besseli1(x::T) where T <: Union{Float32, Float64}
+function besseli1x(x::Float32)
```
oscardssmith, line 71: same here.

src/besseli.jl
heltonmc: Yes good catch! Removing the branch in the Float64 makes it equal to the Float32 now. I have merged them. I also went and merged the scaled functions to match.

heltonmc:
I've recalculated all the coefficients used in these routines. These new weights signficinatly improve errors at large input arguments. I've also added new documentation for internal and external api
 
## `besseli0`
Screenshot from 2022-01-15 15-21-50


## `besseli1`
Screenshot from 2022-01-15 16-25-37

I think this is ready to merge now.

oscardssmith:
One question: Why did you only use 75 digits for Arb? I would recommend using at least 256 since it's basically free.

heltonmc:
I wasn't very clear and for repeatability purposes let me post an example code.
```julia
 bessi1(x) = besseli(1, ArbFloat(x, digits=75))
g(y) = inv(y) * (bessi(2*sqrt(y)) - 1)
N,D,E,X = ratfn_minimax(g, (1e-18, (7.75/2)^2), 6, 0)
```
I just went and reran with digits=256 and they match with digits=75 to about 49 digits.
So for example Ill get something like
```julia
# digits = 75
julia> N
7-element Vector{BigFloat}:
 0.08333336088599108314227728172065098704108098333505801124750081760507945728818461
.
.
# digits = 256
julia> N
7-element Vector{BigFloat}:
 0.0833333608859910831422772817206509870410809833845858298944772410690681516396821
.
.
```

I then will convert them to Float64 or Float32. I think this is ok for Float64. One thing is do we want to also give Float128 or DoubleFloats or higher precision implementations? I do think this is important.... Should we just take on the DoubleFloats or Quadmath dependency and define constants for this? This shouldn't be too hard.

It's also important to document this issue https://github.com/JeffreySarnoff/ArbNumerics.jl/issues/48 with `ArbNumerics.jl`. This makes the evaluation quite slow but should always be careful that your true functions are calculated correctly..

PR #7, opened 2022-01-14
Faster implementations of Besselk0 and Besselk1

heltonmc:
Implementation of `besselk0`, `besselk1` and their scaled versions based on approximations given [here](https://www.advanpix.com/2015/11/25/rational-approximations-for-the-modified-bessel-function-of-the-second-kind-k0-for-computations-with-double-precision/)

The implementation follows similar discussion in #6 by reducing a couple of branches and merging `Float32` and `Float64`.

This comes with quite a performance boost...
```julia
# old version
julia> @benchmark besselk0(x) setup=(x = 100*rand())
 Range (min … max):  15.514 ns … 52.198 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     36.564 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   36.300 ns ±  2.779 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▂                                                    █   ▃ ▁
  ▅█▃▁▁▄▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▄▁▄█ █
  15.5 ns      Histogram: log(frequency) by time      38.3 ns <

# new version
julia> @benchmark besselk0(x) setup=(x = 100*rand())
 Range (min … max):   8.925 ns … 29.659 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     13.693 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.692 ns ±  0.867 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                             ▆█               ▁
  ▅▁▆█▃▃▃▁▁▁▁▁▁▁▁▁▁▃▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃███▃▁▁▁▁▁▁▁▁▁▁▅█ █
  8.92 ns      Histogram: log(frequency) by time      15.2 ns <

# comparing to SpecialFunctions.jl
julia> @benchmark besselk(0, x) setup=(x = 100*rand())
BenchmarkTools.Trial: 10000 samples with 815 evaluations.
 Range (min … max):  122.103 ns … 592.071 ns  ┊ GC (min … max): 0.00% … 65.17%
 Time  (median):     200.658 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   215.430 ns ±  46.820 ns  ┊ GC (mean ± σ):  0.04% ±  0.94%

             ▇▆██▆▇▆▅▆▅▄▄▃▃▃▃▂▂▁▁▁▁                             ▂
  ▂▄▅▅▂▆▆▇▇▇▇█████████████████████████▇▇██▇▇▆▇▇▆▇▅▆▆▅▆▇▇▆▅▆▆▆▅▅ █
  122 ns        Histogram: log(frequency) by time        435 ns <

```

These implementations also handle large arguments better at the edges of float precision
```julia
# SpecialFunctions.jl
julia> besselk(0, 700.0)
0.0

# Bessels.jl
julia> besselk0(700.0)
4.669776431685378e-306
```


oscardssmith:
How do the maximum errors compare?

heltonmc:
Old errors compared to `ArbNumerics.jl`
Screenshot from 2022-01-16 14-09-09

New errors

Screenshot from 2022-01-16 14-06-47

New error average rel. error / eps(Float64) = 0.51424
maximum = 2.1939




Optimizing these polynomials any further is pretty prohibitive due to the issue with `besselk` in `ArbNumerics.jl`. I've commented https://github.com/JeffreySarnoff/ArbNumerics.jl/issues/48. This essentially makes `Remez.jl` take hours to give meaningful results after setting precision above 2000.

heltonmc:
I've added implementations for `besselk(nu,x)` and `besselkx(nu, x)`. These are accurately and quickly calculated with a three term recurrence using `besselk0` and `besselk1`. For small to moderate orders this is ok and most likely the fastest way to do this. 

```julia
# Bessels.jl
julia> @benchmark besselk(50, x) setup=(x=rand()*30)
BenchmarkTools.Trial: 10000 samples with 978 evaluations.
 Range (min … max):  66.969 ns … 108.788 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     72.463 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   72.645 ns ±   1.870 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▃▁                      ▁█▆▆▂     ▃▁▁                 ▂
  ▆▆▃▁▁▁███▆▄▁▃▁▁▅▅▅▁▃▁▁▁▁▄▁▁▁▃▁▃██████▅▅▃████▇▇▆▆▆▆▇██▇▇▆▆▄▆▆ █
  67 ns         Histogram: log(frequency) by time        77 ns <

## Compared to SpecialFunctions.jl
julia> @benchmark besselk(50, x) setup=(x=rand()*30)
BenchmarkTools.Trial: 10000 samples with 200 evaluations.
 Range (min … max):  399.605 ns … 949.175 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     515.200 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   538.276 ns ±  65.815 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

          ▁▁▁▁  ▅█▇▇▆▅▅▄▄▄▃▄▃▃▃▂▃▂▂▂▁▂▁▁▁▁▁▁ ▁                  ▂
  ▅▆▇██████████████████████████████████████████████▇▇▇▇▇▇█▇▇▆▇▇ █
  400 ns        Histogram: log(frequency) by time        792 ns <

```

When the order is large and x is large it is much faster to use a uniform expansion to calculate this. I have added notes in the document about how to implement this. We should come back at a later time to implement this as it requires other bessel functions we haven't implemented yet. 

```julia
# Bessels.jl
julia> @benchmark besselk(120, x) setup=(x=rand()*1000 + 10.0)
 Range (min … max):   27.027 ns …   5.351 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     137.620 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   288.813 ns ± 879.224 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆█                                                          ▂ ▁
  ██▇▃▁▁▁▁▁▁▃▃▁▁▁▃▄▃▁▁▄▃▄▄▁▄▄▅▁▄▄▃▅▃▁▄▄▄▄▄▄▅▄▃▄▄▄▄▅▄▅▅▅▄▅▅▄▄▄▅█ █
  27 ns         Histogram: log(frequency) by time       4.79 μs <

# SpecialFunctions.jl
julia> @benchmark besselk(120, x) setup=(x=rand()*1000 + 10.0)
BenchmarkTools.Trial: 10000 samples with 631 evaluations.
 Range (min … max):  194.799 ns …  1.524 μs  ┊ GC (min … max): 0.00% … 86.15%
 Time  (median):     207.323 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   206.982 ns ± 19.862 ns  ┊ GC (mean ± σ):  0.13% ±  1.22%

   █                   █▇                                       
  ▄█▂▂▂▃▅▅▂▃▅▃▃▃▃▂▂▂▂▂▃██▇▆▂▅▇▆▄▇▅▄▄▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▁▂▃▂▁▁▂▂▂▂ ▃
  195 ns          Histogram: frequency by time          229 ns <

```

It looks like `SpecialFunctions.jl` switches over to this expansion with the amos routines. Which becomes faster starting at around order 200.

@oscardssmith Could you review this three term recurence function? Its speed is pretty important for fast generation of these higher order terms. It will also be used for the `bessely(nu, x)` implementation. 

One thing we should consider is returning an array of nu values. This is usually how you need them anyway so we should also have an implementation that returns the array.... will have to think about best way to do this.

src/besselk.jl
```diff
@@ -114,3 +129,46 @@ function besselk1x(x::T) where T <: Union{Float32, Float64}
...
+@inline function three_term_recurrence(x, k0, k1, nu)
+    nu == 0 && return k0
+    nu == 1 && return k1
+
+    # this prevents us from looping through large values of nu when the loop will always return zero
+    (iszero(k0) || iszero(k1)) && return zero(x) 
+
+    k2 = k0
+    for n in 1:nu-1
+        a = 2 * n / x
```
oscardssmith, line 168: Save `2/x` Division is expensive.

src/besselk.jl
```diff
@@ -114,3 +129,46 @@ function besselk1x(x::T) where T <: Union{Float32, Float64}
...
+        k2 = muladd(a, k1, k0)
```
oscardssmith, line 170: Will `a*k1>k0` generally hold? If not, this should probably use compensated arithmetic to ensure an accurate answer.

heltonmc: It doesn't. when x is large k0 will be bigger.

heltonmc: This does seem to matter
```julia
# ArbNumerics
julia> Float64(besselk(500, ArbFloat(100)))
2.7313831719901783e279
# current implementation in BigFloat
julia> Float64(three_term_recurrence(BigFloat(100), BigFloat(k0), BigFloat(k1), 500))
2.7313831719901783e279
# current implementation in Float64
julia> besselk(500, 100.0)
2.731383171990204e279
```
Though I am not sure the best way to compensate for this........

oscardssmith: Does this work?
```
@inline function three_term_recurrence(x::T, k0, k1, nu) where T<:Number
    nu == 0 && return k0
    nu == 1 && return k1

    # this prevents us from looping through large values of nu when the loop will always return zero
    (iszero(k0) || iszero(k1)) && return zero(x) 

    k2 = k0
    k0lo = k1lo = k2lo = zero(T)
    invx2 = 2 / x
    invx2lo = fma(invx2, -x, 2) / x
    for n in 1:nu-1
        a = invx2 * n
        alo = invx2lo * n
        k2 = a * k1
        k2lo = fma(a, -k1, k2) + muladd(k1, alo, muladd(a, k1lo, k0lo))
        k2new = k2 + k0
        k2lo += k2 - (k2new - k0)
        k2 = k2new
        k0, k0lo = k1, k1lo
        k1, k1lo = k2, k2lo
    end
    return k2 + k2lo
end
```

heltonmc: Yes!!!
```julia
julia> three_term_recurrence(100.0, k0, k1, 500)
2.7313831719901783e279
```

heltonmc: Though there is a performance hit to doing this...

```julia
## your version
julia> @benchmark besselk(200, x) setup=(x=rand()*1000 + 20.0)
BenchmarkTools.Trial: 2481 samples with 995 evaluations.
 Range (min … max):   27.042 ns … 27.604 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     769.602 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):     2.024 μs ±  5.500 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆█                                               ▁            
  ██▅▃▁▁▄▁▃▁▁▁▁▃▁▄▃▁▅▁▅▁▁▃▃▁▅▃▄▃▅▁▃▅▃▃▁▃▁▃▃▁▄▁▅▄▃▃▅██▆▅▆▇▆▅▆▅▆ █
  27 ns         Histogram: log(frequency) by time      26.1 μs <

## old version
julia> @benchmark besselk(200, x) setup=(x=rand()*1000 + 20.0)
 Range (min … max):   27.022 ns … 7.930 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     212.084 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   366.389 ns ± 1.151 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █                                                           
  ██▁▁▁▂▂▁▂▁▂▁▂▂▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  27 ns           Histogram: frequency by time        7.92 μs <

```

oscardssmith: I'm honestly kinda shocked I got the error compensation right. That never happens the first time.

heltonmc: Definitely impressed!!

heltonmc:
Ok I am going to merge this and open a new PR with the other method I am working on for higher orders. It is very accurate even for lowish orders. We can move the discussion of the recurrence errors there as well to decide how we want to deal with those. We will have to determine the right asymptotic cutoff but I think this discussion will be best as its own PR. It works for both `besseli(nu, x)` and `besselk(nu, x)`

Issue #9, opened 2022-01-22
Poor accuracy for middle orders 2-50 for besseli


heltonmc:
For orders between 2-50 in `besseli`, SpecialFunctions.jl gives superior accuracy.

Summary: High accuracy is found for `nu = 0,  1`. But for `nu` between 2-40 poor accuracy is achieved. This is due to using the asymptotic expansion for large expansion and then doing down recurrence. We are limited by the errors in the asymptotic expansion. For `nu=40`, the accuracy is matched except for low arguments. It appears the fortran routines have a cutoff for different values of x. I am assuming they are using an expansion for large arguments when `x>>nu` to achieve better accuracy. After `nu>80` accuracy is comparable. Probably because a similar asymptotic expansion is used for all arguments. 

Solution: We should (unfortunately) create another branch for medium orders. The two best ways either are Miller's algorithm (fortran routines uses) or a continued fraction approach. I have a continued fraction approach going now that can achieve excellent accuracy. I'll do this as a PR and we can then do some benchmarks. Again, determining where we want these cutoff to be will be a matter of preference of speed and accuracy..... ideally this would be as smooth as possible.

# nu = 0  - Good accuracy
![nu0](https://user-images.githubusercontent.com/57471570/150657749-40737fda-66a8-4f37-bcfc-c83a72a5577d.png)

# nu = 10 - Bad accuracy
![nu10](https://user-images.githubusercontent.com/57471570/150657755-153b877b-3312-4a29-b9ab-83795c57568a.png)

# nu = 40
![nu40](https://user-images.githubusercontent.com/57471570/150657758-4f9c606e-58b6-4b52-a2a4-5bdc3804bbb1.png)

# nu = 80
![nu80](https://user-images.githubusercontent.com/57471570/150657761-19c8f328-6b08-48ec-a890-0f730966843b.png)

# nu = 120
![nu120](https://user-images.githubusercontent.com/57471570/150657765-0ebd0031-d3d2-41e4-b45e-e11af2f4881a.png)

# nu = 250
![nu250](https://user-images.githubusercontent.com/57471570/150657767-54e3efd4-557b-40da-8031-bfcb4a508784.png)

heltonmc:
Some comparable benchmarks.

```julia
julia> @benchmark besseli(40, x) setup=(x=rand()*100)
BenchmarkTools.Trial: 10000 samples with 905 evaluations.
[...]

PR #10, opened 2022-01-22
Use continued fractions for besseli for medium sized orders 2<nu<100

heltonmc:
This attempts to solve #9 by implementing a continued fractions approach. Here is the same plot shown in that issue for `nu=40` which is the problem region. 
nu40_steed

This algorithm can generate very accurate results. The computation time is what should be discussed. The main challenge is quickly determining the continued fraction. I am using Steed's algorithm to do this. It still is quite fast.

```julia
julia> @benchmark Bessels._besseli_continued_fractions(40, x) setup=(x=rand()*50)
BenchmarkTools.Trial: 10000 samples with 947 evaluations.
 Range (min … max):   87.985 ns … 263.846 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     198.003 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   193.414 ns ±  41.931 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                   ▂  ▅  ▅  ▅  ▄  ▅  ▄      ▅  █  ▇  ▆  ▄  ▅     
  ▂▁▁▃▁▁▅▁▁▂▆▃▁▆▆▁▄█▂▁█▂▁█▂▁█▃▂█▄▁█▄▂█▅▂██▂▄█▂▂█▂▁█▃▂█▄▁█▆▂█▆▂█ ▄
  88 ns            Histogram: frequency by time          261 ns <

julia> @benchmark SpecialFunctions.besseli(40, x) setup=(x=rand()*50)
BenchmarkTools.Trial: 8421 samples with 951 evaluations.
 Range (min … max):   90.997 ns …   1.985 μs  ┊ GC (min … max): 0.00% … 59.25%
 Time  (median):     779.588 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   623.006 ns ± 294.210 ns  ┊ GC (mean ± σ):  0.04% ±  1.18%

    ▁▁                                                ▃█▃▅▃      
  ▃▇██▃▂▁▁▁▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃█████▅▃▄▃ ▃
  91 ns            Histogram: frequency by time          871 ns <

```

But it is significantly slower than the asymptotic expansion. It also is not independent of x. It takes longer to converge for large arguments of x.... but maybe we can think of doing an asymptotic expansion there for large arguments.

@oscardssmith the function `steed` is the performance killer here. I don't like checking for convergence at each loop. Virtually all of them take between 15-20 iterations to converge except at very large arguments...

All in all this method can deliver very accurate results at moderate speeds. It is still significantly faster than `SpecialFunctions.jl` but slower than the asymptotic expansion (1.5x - 2x).


oscardssmith:
Can you hard-code the first 10 iterations as a ratio of 2 polynomials? If not, is there a better way you can initialize `a` and `h` such that you'll get better convergence?

heltonmc:
I'll have to investigate further...... There are a few algorithms to solve this I think. And to be clear we are trying to solve the continued fraction of form...
```julia
        1                       1                              1
------------ -----------------  -----------------
2*nu*inv(x)+ 2*(nu+1)*inv(x)+   2*(nu+2)*inv(x)
```

oscardssmith:
What's with the formatting here?

heltonmc:
Ya that's a bad format. All it's doing is this
```julia

nu = 10
x = 10.0
xinv = inv(x)

a13 = 2*(nu + 15)*xinv
a12 = 2*(nu + 14)*xinv + inv(a13)
a11 = 2*(nu + 13)*xinv + inv(a12)
a10 = 2*(nu + 12)*xinv + inv(a11)
a9 = 2*(nu + 11)*xinv + inv(a10)
a8 = 2*(nu + 10)*xinv + inv(a9)
a7 = 2*(nu + 9)*xinv + inv(a8) 
a6 = 2*(nu + 8)*xinv + inv(a7)
a5 = 2*(nu + 7)*xinv + inv(a6)
a4 = 2*(nu + 6)*xinv  + inv(a5)
a3 = 2*(nu + 5)*xinv  + inv(a4)
aa = 2*(nu + 4)*xinv + inv(a3)
a = 2*(nu + 3)*xinv + inv(aa)
b = 2*(nu + 2)*xinv + inv(a)
c = 2*(nu + 1)*xinv + inv(b)
d = 2*nu * xinv + inv(c)

h = inv(d)
```

But keeping track of error terms along the way to see when it converges.

oscardssmith:
What about putting a bound on the number of iterations you need (either theoretically or empirically), and just doing 5 more than that unconditionally without checking for convergence?

heltonmc:
So the good thing is we can place fairly tight bounds on our problem. We are looking at `2<nu<100` and `0<x<700`.
Higher orders and smaller arguments converge faster. Though, large x in general is the slowest. So `nu=2` and `x=700` is the worst taking 157 iterations to converge whereas for `n=5` and `x=10` only 17 iterations are needed.... so unfortunately a large spread

The checks are fairly expensive too... the no check version with a fixed loop size is around 1.5x faster.

oscardssmith:
One thing you could always do is you could sample something like 10^8 random combinations of `nu` and `x` (which would take about 20 seconds), and see if you can find a function that fits it well. If the no-check version is 1.5x faster, you get a decent amount of allowed over-estimation of required checks before you lose out.

heltonmc:
Ok so I have looked into the problem a little bit more. The continued fractions approach is very accurate for all regions of x and nu. Though it gets very slow for large orders (has to use forward recursion to get there) and very slow for large arguments. The large orders is solved by using uniform asymptotic expansion in regards to `nu`.

Now, the large argument problem can be solved by just switching over to Hankel's asymptotic expansion. https://dlmf.nist.gov/10.40 (10.40.1). A quick implementation of this is pretty straightforward and looks like:
```julia
function besi(v, z::T) where T
    coef = exp(z) / sqrt(2*T(pi)*z)
    a1 = (4v^2 - 1) / 8 / z
    a2 = (4v^2 - 1)*(4v^2 - 9) / 8^2 / z^2 / 2
    a3 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25) / 8^3 / z^3 / 6
    a4 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49) / 8^4 / z^4 / 24
    a5 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81) / 8^5 / z^5 / 120
    a6 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121) / 8^6 / z^6 / 720
    a7 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169) / 8^7 / z^7 / 5040
    a8 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225) / 8^8 / z^8 / 40320
    a9 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289) / 8^9 / z^9 / 362880
    a10 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361) / 8^10 / z^10 / 3628800
    a11 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441) / 8^11 / z^11 / 39916800
    a12 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441)*(4v^2 - 529) / 8^12 / z^12 / 479001600
    a13 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441)*(4v^2 - 529)*(4v^2 - 625) / 8^13 / z^13 / 6227020800
    a14 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441)*(4v^2 - 529)*(4v^2 - 625)*(4v^2 - 729) / 8^14 / z^14 / 87178291200
    a15 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441)*(4v^2 - 529)*(4v^2 - 625)*(4v^2 - 729)*(4v^2 - 841) / 8^15 / z^15 / 1307674368000
    a16 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441)*(4v^2 - 529)*(4v^2 - 625)*(4v^2 - 729)*(4v^2 - 841)*(4v^2 - 961) / 8^16 / z^16 / 20922789888000
    a17 = (4v^2 - 1)*(4v^2 - 9)*(4v^2 - 25)*(4v^2 - 49)*(4v^2 - 81)*(4v^2 - 121)*(4v^2 - 169)*(4v^2 - 225)*(4v^2 - 289)*(4v^2 - 361)*(4v^2 - 441)*(4v^2 - 529)*(4v^2 - 625)*(4v^2 - 729)*(4v^2 - 841)*(4v^2 - 961)*(4v^2 - 1089) / 8^17 / z^17 / 355687428096000

    return coef*(1 - a1 + a2 - a3 + a4 - a5 + a6 - a7 + a8 - a9 + a10 - a11 + a12 - a13 + a14 - a15 + a16 - a17)
end
```

Obviously that can be optimized better for performance. The real question is appropriately choosing cutoff points which balances the number of terms needed and your minimum cutoffs.
So far it looks like as long as `z > max{30.0, nu^2 / 2}` this expansion will work. Now if we increased the limits to where we apply this we could potentially use less terms. Have to test performance etc. to see how costly extra terms are compared to the continued fraction method. Both of these methods should produce very low errors assuming right cutoffs and number of terms

oscardssmith:
Here's the cleaned up version. (it should also be faster).
```
function besi(v, z::T) where T
    coef = exp(z) / sqrt(2*T(pi)*z)
    fv2 = 4v^2
    term = one(T)
    res = term
    s = -term
    for i in 1:17
        offset = muladd(2, i, -1)
        term *= 1/8*(muladd(offset, -offset, fv2)/(z*i)
        res = muladd(term, s, res)
        s = -s
    end
    return res*coef
end
```

heltonmc:
Ya that is sooooooo fast it hardly matters if we add more terms to the series. I'm going to more rigorously test the limits of where this expansion can be applied and go from there. I believe at that point we will have rounded out all of `besselk` and `besseli` for all integer orders and real arguments (though these routines also work for non integer and complex arguments just have to test for that).

oscardssmith:
It turns out that turning a quadratic algorithm into a linear one is good for performance (shocking, I know).

heltonmc:
Great! I will add in the check and go from there. I was noticing some of the tests are failing for `Float32` inputs.

```julia
besseli(2, 8.6) = 586.4199234124159
besseli(2, 8.6f0) = 586.42017f0
```

Though this happens when the `Float32` can't be exactly represented by `Float64`. Internally, the code is coverting it to a Float 64... `Float64(8.6f0)=8.600000381469727` before doing computation which yields the wrong answer. Is there a way to ensure this conversion is more exact?

oscardssmith:
All `Float32` can be exactly represented by `Float64`. The problem is that `Float64(Float32(x)) != Float64(x)`. The solution is that when testing `Float32` values, you should make sure that the values you test are exactly representable as `Float32`.

heltonmc:
Ok I think I've fixed things to where everything is working. There are a few things I'm not happy with and will try to clean up. Almost all of the internal calculations need to happen in `Float64` even for `Float32` arguments. The differences that need to happen are that for `Float32` we can use a lower amount of U-polynomials and in the continued fractions algorithm we only need to iterate until `error<eps(Float32)`. This is the same for the large_argument branch. 

The issue with the continued fractions is it uses upward recurrence from `besselk`. Those starting values can very easily return 0 in Float32 if you aren't careful but the `besseli` calculation at higher orders can be in range. So we should also recover those values in `Float64`. I think this can better be done in the algorithm but will have to work through it probably later today or tomorrow.

heltonmc:
Since the continued fraction algorithm is very slow I have been investigating potentially other ways to calculate this better. For medium orders and arguments (`nu=50`, `x=50`) there's not really a good way to solve this other than the continued fractions algorithm here or Miller's algorithm (though some of the papers mention the continued fractions is more efficient.... I think that's left to be seen).

Anyway for smallish arguments or when `x` is somewhat less than `nu` it turns out just using the power series definition is rapidly convergent and can be coded quite fast using recurrence. Here is an implementation.

```julia
function besseli_small_arguments(v, z::T) where T
    if v < 20
        coef = (z / 2)^v / factorial(v)
    else
        coef = v*log(z / 2)
        coef -= loggam(v + 1)
        coef = exp(coef)
    end

    MaxIter = 200
    out = one(T)
    zz = z^2 / 4
    a = one(T)
    for k in 0:MaxIter
        a *= zz / (k + 1) / (k + v + 1)
        out += a
        a <= eps(T) && break
    end
    return coef * out
end
@inline function loggam(x)
    xinv = inv(x)
    xinv2 = xinv * xinv
    out = (x - 0.5) * log(x) - x + 9.1893853320467274178032927e-01
    out += xinv * evalpoly(
        xinv2, (8.3333333333333333333333368e-02, -2.7777777777777777777777776e-03,
        7.9365079365079365079365075e-04, -5.9523809523809523809523806e-04)
    )
    return out
end
```
The trick is calculating `(x/2)^v / v!` accurately. I essentially just used a couple of terms of Sterling's approximation for log(x!) to get this (or loggama). It is convergent for orders over 20. Though, it is faster and more accurate to just use the factorial lookup for smaller orders.

Let me know if you see anything to improve with this function or better ways to handle the factorials.

src/besseli.jl
```diff
@@ -193,3 +192,48 @@ function besseli_large_orders_scaled(v, x::T) where T <: Union{Float32, Float64,
...
+function besseli_large_argument(v, z::T) where T
+    MaxIter = 1000
+    S = promote_type(T, Float64)
+    z = S(z)
+    coef = exp(z) / sqrt(2*S(pi)*z)
+    fv2 = 4v^2
+    term = one(S)
+    res = term
+    s = -term
+    for i in 1:MaxIter
```
oscardssmith, line 230: It might make sense to make `i` be of type `T`. Not sure if it will change anything, but it might improve performance.

src/besseli.jl
```diff
@@ -136,17 +136,16 @@ end
...
 """
-function besseli(nu, x::T) where T <: Union{Float32, Float64, BigFloat}
+function besseli(nu, x::T) where T <: Union{Float32, Float64}
     nu == 0 && return besseli0(x)
     nu == 1 && return besseli1(x)
-
-    branch = 60
-    if nu < branch
-        inp1 = besseli_large_orders(branch + 1, x)
-        in = besseli_large_orders(branch, x)
```
oscardssmith, line 146: don't use `in` as a variable name.

oscardssmith:
I think you should still look at optimizing the continued fraction approach a little more. Since `besseli_large_argument` takes care of the case where you need a lot of evals of the fraction, I think estimating the number of iterations is probably the better approach. It should be a free speedup.

heltonmc:
I'll investigate further! It definitely takes care of the hardest cases, but there are still a few problem areas (largish orders) I'll look into. Here's a sample of where we probably have to use the continued fractions and the number of iterations required to converge. 
```julia
julia> steed(60, 600.0)
count = 102

julia> steed(40, 250.0)
count = 65

julia> steed(2, 15.0)
count = 24
```

src/besseli.jl
heltonmc: This does slightly improve performance.

src/besseli.jl
```diff
@@ -193,3 +192,48 @@ function besseli_large_orders_scaled(v, x::T) where T <: Union{Float32, Float64,
...
+    MaxIter = 1000
+    xinv = inv(x)
+    xinv2 = 2 * xinv
+    d = x / (n + n)
+    a = d
+    h = a
+    b = muladd(2, n, 2) * xinv
+    for _ in 1:MaxIter
+        d = inv(b + d)
+        a = muladd(b, d, -1) * a
```
oscardssmith, line 213: This can be
```
d_new = inv(b+d)
a *= -d/d_new
d = d_new
```
I think this should be slightly faster

oscardssmith:
Oh, I just looked at `steed` a little more, and there's an easy optimization. If you can get a guess of `h`, you don't have to compute `a/h`, but instead compute `a*h_guess` for your convergence check.

heltonmc:
So I went ahead and tried the other popular approach (modified lentz algorithm). Here it is.
```julia
function modified_lentz(v, x::T) where T
    MaxIter = 500
    f = 1e-30
    c = f
    d = zero(x)
    xinv = inv(x)
    b1 = 2 * v * xinv
    b = b1

    for i in 1:MaxIter
        d += b
        c = b + inv(c)
        d = inv(d)
        delta = c * d
        f *= delta
        b = b1 + 2 * i * xinv
        abs(delta - 1) < eps(T) && break
    end
    return f
end
```
Should return same value as `steed`.... on some quick benchmarks though it appears `steed` was slightly faster though the convergence check here is much simpler. Really just want to know when `isone(delta)`....

What do you think on these two algorithms?

oscardssmith:
The initialization for `f` here is very suspect. Why is it that value in particular?

heltonmc:
There is a danger in the original method of Lent's that if b0 is zero which it is here it invalidates the recurrence relation. It is then posed from https://www.sciencedirect.com/science/article/pii/002199918690046X to modify this by shifting the starting value to some tiny value below machine precision and therefore avoid the division by zero. In fact you always have to check if `c` and `d` are zero and shift them accordingly but for our application it is only necessary at the beginning.

oscardssmith:
If, for parameters `nu, x` you use `1.2*(6.018*exp(-0.0494*nu)+0.4241)*x^(0.003612*nu+0.5325)` iterations, you will always use enough iterations to converge, and be able to evaluate backwards rather than forward, so it will be faster.

oscardssmith:
I've played with this a little more, and `iters(nu, x)=1.25*(exp2(4.82*(exp2(-.0267*nu))-2.38)*x^(.89-3/8*exp2(-.0366*nu))+5)` is the function we want to compute needed iterations. On average it uses 25% too many iterations, and 2^20 random tests don't show a case where it uses too few iterations.

heltonmc:
Awesome! The algorithm can be simplified too if we know the end term. So comparing....

```julia
function steed(n, x::T) where T
    MaxIter = 150
    xinv = inv(x)
    xinv2 = 2 * xinv
    d = x / (n + n)
    a = d
    h = a
    b = muladd(2, n, 2) * xinv
    #count = 0
    for _ in 1:MaxIter
        d = inv(b + d)
        a = muladd(b, d, -1) * a
        h = h + a
        b = b + xinv2
        abs(a / h) <= 1e-200 && break
        #count += 1
    end
    #@show count
    return h
end
function steed1(nu, x::T) where T
    MaxIter = 150#iters(nu, x)
    #@show MaxIter
    xinv = inv(x)

    a = 2 * (nu + MaxIter) * xinv
    for i in MaxIter-1:-1:0
        a = 2 * (nu + i) * xinv + inv(a)
    end
    return inv(a)
end

julia> @benchmark steed(x, y) setup=(x=19; y = 60.0)
 Range (min … max):  1.400 μs …  4.046 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.413 μs              ┊ GC (median):    0.00%

julia> @benchmark steed1(x, y) setup=(x=19; y = 60.0)
BenchmarkTools.Trial: 10000 samples with 190 evaluations.
 Range (min … max):  527.411 ns … 811.842 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     528.289 ns               ┊ GC (median):    0.00%
```

So there is a definite performance gain if you know the end point and don't have to check. Now I think the challenge is still getting it right. For an example I tried it was getting too low iterations.... I'll look into this some more
```
julia> steed(19, 60.0)
count = 128
0.7363674261364371

julia> steed1(19, 60.0)
MaxIter = 43.57873177521482
0.7295762234926214
```

oscardssmith:
I was basing this off of `modified_lentz` which only needs 35 iterations. Why do the different algorithms need different numbers of iterations?

heltonmc:
No I forgot to change the break condition it was too low. The number of iterations should be similar though it looks like `steed` has consitently two less in most of the benchmarks.
```julia
julia> modified_lentz(5, 10.0)
count = 19
0.633668391623305

julia> steed(5, 10.0)
count = 17
0.6336683916233057

julia> steed(19, 60.0)
count = 33
0.7363674261364371

julia> modified_lentz(19, 60.0)
count = 35
0.7363674261364366
```

oscardssmith:
How does `steed1` compare to the obvious evaluation scheme (of just computing from right to left)?

heltonmc:
These benchmarks are starting to get noisy. Ill have to do this on my desktop later to see. But there doesn't seem to be clear cut.

```julia
julia> @benchmark steed1(x, y) setup=(x=19; y = rand() + 60.0)
BenchmarkTools.Trial: 10000 samples with 298 evaluations.
 Range (min … max):  281.601 ns … 570.748 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     376.819 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   375.013 ns ±  27.864 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

julia> @benchmark steed1(x, y) setup=(x=19; y = rand() + 120.0)
BenchmarkTools.Trial: 10000 samples with 204 evaluations.
 Range (min … max):  378.265 ns … 766.951 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     464.049 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   461.086 ns ±  26.422 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
```

oscardssmith:
One free 3% speedup in `steed1` is using `twoinvx = 2/x` instead of `invx`, and `muladd(nu+i, twoinvx, inv(res))` in the loop.

oscardssmith:
Is this ready to merge?

heltonmc:
Yes I think so. There are three things that may be worth considering separately in issues/PRs.

1. The steed algorithm seems to perform better than the modified lentz and there doesn't seem to be a clear advantage to having the number of iterations needed unless for really large arguments. For example,
```julia
function steed(n, x::T) where T
    MaxIter = 1000
    xinv = inv(x)
    xinv2 = 2 * xinv
    d = x / (n + n)
    a = d
    h = a
    b = muladd(2, n, 2) * xinv
    for _ in 1:MaxIter
        d = inv(b + d)
        a = muladd(b, d, -1) * a
        h = h + a
        b = b + xinv2
        abs(a / h) <= eps(T) && break
    end
    return h
end
function steed1(nu, x::T) where T
    MaxIter = 1.25*(exp2(4.82*(exp2(-.0267*nu))-2.38)*x^(.89-3/8*exp2(-.0366*nu))+5)
    xinv2 = 2 * inv(x)
    a = (nu + MaxIter) * xinv2
    for i in MaxIter-1:-1:0
        a = muladd(xinv2, (nu + i), inv(a))
    end
    return inv(a)
end

julia> @benchmark steed(x, y) setup=(x=19;y=rand()+10.0)
BenchmarkTools.Trial: 10000 samples with 939 evaluations.
 Range (min … max):  102.946 ns … 161.741 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     104.411 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   105.468 ns ±   3.195 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▁  █▇          ▄▂ ▂▃           ▁  ▁                         ▂
  ▇██████▇▇███▇█▇█▇█████▇█████████▇██████▇▅▆▅▅▅▅▄▄▄▅▄▄▄▄▃▄▄▁▄▃▅ █
  103 ns        Histogram: log(frequency) by time        118 ns <

julia> @benchmark steed1(x, y) setup=(x=19;y=rand()+10.0)
BenchmarkTools.Trial: 10000 samples with 690 evaluations.
 Range (min … max):  166.545 ns … 551.752 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     265.580 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   265.926 ns ±  29.475 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                        ▁   ▄   ▇ ▂▇ ▄ █   ▆ ▁ ▂                 
  ▂▁▂▁▁▂▂▂▂▂▃▂▃▁▄▂▄▂▆▃▂▅█▂▇▂█▄█▂█▄██▄█▃█▅▃██▃█▂█▄▆▇▃▃▄▂▆▃▃▂▃▃▃▃ ▄
  167 ns           Histogram: frequency by time          336 ns <

```
So for some arguments it is considerably slower.

2. The use of the the power series for small arguments. I am going to leave the code for this in the file but power series converges very fast for smal larguments. The challenge is when `nu` is larger than the factorial lookup we get a decrease in the numerical accuracy. We have to calculate `(x/2)^nu / factorial(nu)` accurately but the way I am currently doing it leads to errors. Copied below is the relevant code...
```julia
if v < 20
        coef = (z / 2)^v / factorial(v)
    else
        coef = v*log(z / 2)
        coef -= loggam(v + 1)
        coef = exp(coef)
    end
```
3. We should consider how we want to implement the scaled versions of these functions. It might make sense to have the scaled versions be the base code and have the regular functions just call the scaled functions with an exponential factor. This will decrease the amount of code as most of it is just copy pasted.

Again it might be worth considering all of these separately as the functions right now are very accurate and still much faster than SpecialFunctions.jl.

oscardssmith:
Can you use the series expansions of `exp(a*nu)/factorial(nu)` where `a=log(x/2)` (see https://www.wolframalpha.com/input?i=exp%28a*x%29+%2F+factorial%28x%29). The expansion around `nu=Inf` looks like it would work really well.

src/besseli.jl
```diff
@@ -193,3 +194,83 @@ function besseli_large_orders_scaled(v, x::T) where T <: Union{Float32, Float64,
...
+        coef -= _loggam(v + 1)
+        coef = exp(coef)
+    end
+
+    MaxIter = 1000
+    out = one(T)
+    zz = z^2 / 4
+    a = one(T)
+    for k in 0:MaxIter
+        a *= zz / (k + 1) / (k + v + 1)
```
oscardssmith, line 261: `zz / ((k + 1)  * (k + v + 1))` should be faster.

src/besseli.jl
```diff
@@ -193,3 +194,83 @@ function besseli_large_orders_scaled(v, x::T) where T <: Union{Float32, Float64,
...
+end
+
+# power series definition of besseli
+# fast convergence for small arguments
+# for large orders and small arguments there is increased roundoff error ~ 1e-14
+# need to investigate further if this can be mitigated
+function besseli_small_arguments(v, z::T) where T
+    if v < 20
+        coef = (z / 2)^v / factorial(v)
+    else
```
oscardssmith, line 250:
```
vinv = inv(v)
coef = sqrt(inv(v)*2π) * e^(v (log(z/(2*v)) + 1)) * evalpoly(vinv, (1, -1/12, 1/288,  139/51840, -571/2488320))
```

heltonmc:
Ya that looks promising! Looking at this function now
```julia
function t(v, z::T) where T
    vinv = inv(v)
    coef = sqrt(vinv/2π) * MathConstants.e^(v*(log(z/(2*v)) + 1)) 
    coef *= evalpoly(vinv, (1, -1/12, 1/288,  139/51840, -571/2488320, -163879/209018880, 5246819/75246796800, 534703531/902961561600))


    MaxIter = 1000
    out = one(T)
    zz = z^2 / 4
    a = one(T)
    for k in 0:MaxIter
        a *= zz / (k + 1) / (k + v + 1)
        out += a
        a <= eps(T) && break
    end
    return coef * out
end
```
We can get very good results even for moderate values of `nu`
```julia
julia> v = 30
30

julia> x = 10.0
10.0

julia> Float64(1 - t(v, x) / besseli(v, ArbFloat(x)))
-7.891895011035064e-16
```
Though for larger values of `nu` we do get some loss in accuracy
```julia
julia> x = 10.0
10.0
julia> v = 200
200

julia> Float64(1 - t(BigFloat(v), BigFloat(x)) / besseli(v, ArbFloat(x)))
-1.951401549180306e-17

julia> Float64(1 - t(v, x) / besseli(v, ArbFloat(x)))
-2.8298868165059427e-14

**Comment by heltonmc** [comment]
Though, I guess we switch over to the asymptotic expansion then so the maximum error would be
```julia
julia> v = 99
99
julia> x = 10.0
10.0

julia> Float64(1 - t(v, x) / besseli(v, ArbFloat(x)))
7.522980473475217e-15

julia> Float64(1 - t(BigFloat(v), BigFloat(x)) / besseli(v, ArbFloat(x)))
-1.9544133749220085e-17
```
Which is slightly better than the current version
```julia
julia> Float64(1 - besseli_small_arguments(v, x) / besseli(v, ArbFloat(x)))
2.0690834071802747e-14
```

src/besseli.jl
heltonmc: I think this block can just be replaced by
```julia
zz = z^2 / 4
    a = one(T)
    for k in 1:MaxIter
        a *= zz / (k * (k + v))
        out += a
        a <= eps(T) && break
    end
```

oscardssmith:
Nice! Is this now ready to merge?

heltonmc:
Just adding in the scaled versions right now and then ready to go!

heltonmc:
There is probably some room to decrease the amount of code overlap between the scaled and unscaled versions. We just need to be careful on over/under flow for these arguments if we combine them.

oscardssmith (approved):

oscardssmith:
Should the readme be updated with new benchmarks?

heltonmc:
Yes absolutely. The readme needs a rewrite at this point as so much has changed. Let me open a separate PR with a benchmark suite so we can figure out best (objective) way to compare.

It might also be worth setting up CI to measure performance changes for PRs? Might help with development to have performance benchmarks automated....

PR #13, opened 2022-02-16
Better large order approximations for J0,J1, Y0,Y1

heltonmc:
So there is some error for the current implementations of these functions.... Here is absolute error for `besselj0`
Screen Shot 2022-02-16 at 12 37 22 PM

A better method is to use expansion based on just cosines as given by Harrison, John. "Fast and accurate Bessel function computation." 2009 19th IEEE Symposium on Computer Arithmetic. IEEE, 2009. Which doesn't rely on the difference between sin and cosines which should be more accurate and faster.

They are indeed faster...
```julia
julia> @benchmark besselj0(x) setup=(x=rand()*5000)
 Range (min … max):   9.086 ns … 53.560 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     29.815 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   29.875 ns ±  1.085 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                             ▁▁      █▇▂▁▁▂▃▁ ▂
  ▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▃▁██▄▄▅▁▁█████████ █
  9.09 ns      Histogram: log(frequency) by time      32.6 ns <

julia> @benchmark besselj0(x) setup=(x=rand()*5000)
 Range (min … max):   7.799 ns … 42.042 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     17.184 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.391 ns ±  1.530 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                              ▇█▄   ▂▁▂                     ▁ ▁
  ▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁███▁▁▄███▁▁▁▁▄▅▅▃▄▁▁▃▁▃▃▃▄▅▄▃▇█ █
  7.8 ns       Histogram: log(frequency) by time      26.5 ns <

```
Which I think is highly advantageous considering they are used often in Hankel transforms etc (integrate from 0 to infinity for x).
However, as article suggests these should be more accurate... . but first implementation doesn't improve accuracy
Screen Shot 2022-02-16 at 12 43 13 PM

Now the issue is the line `b = cos(x + xn)` which leads to bad errors... if we replace that with `b = cos(x)*cos(xn) - sin(x)*sin(xn)` we get errors that look like...
Screen Shot 2022-02-16 at 12 44 36 PM

So we can see that now that is very accurate for that expansion. However, that is significantly slower....


src/besselj.jl
```diff
@@ -31,6 +51,19 @@ function besselj0(x::Float64)
...
+        x2 = xinv*xinv
+
+        p = (one(T), -1/16, 53/512, -4447/8192, 5066403/524288)
+        p = evalpoly(x2, p)
+        a = SQ2OPI(T) * sqrt(xinv) * p
+
+        q = (-1/8, 25/384, -1073/5120, 375733/229376, -55384775/2359296)
+        xn = fma(xinv, evalpoly(x2, q), - PIO4(T))
+        b = cos(x)*cos(xn) - sin(x)*sin(xn)#cos(x + xn)
+        #b = cos(x + xn)
```
heltonmc, line 65: These are the two lines that need some care. `b = cos(x)*cos(xn) - sin(x)*sin(xn)` is very accurate but slow.... 
`b = cos(x + xn)` is fast but less accurate

oscardssmith: Can you rephrase using `cospi`/`sinpi`? Those work better for big inputs.

heltonmc: Hmmm ya I think this needs some type of reduction. The issue is that `x` is really large and `xn` is small and negative so we lose precision making that addition before the `cosine`. Whereas `cos(x)*cos(xn)` makes that reduction for us.

Do you mean like `b = cospi(x/pi + xn/pi)`? Or something different

heltonmc:
```julia
c1 = sincos(x)
c2 = sincos(xn)
b = c1[2]*c2[2] - c1[1]*c2[1]
```
This is just as fast as the old versions and more accurate.... but still 80% slower than `b = cos(x + xn)`

heltonmc:
Ok so I've updated `besselj0`, `besselj1`, `bessely0`, and `bessely1` to use `sincos` when applicable and the different expansion for `cos(x + xn)`.
The new absolute errors look like this now...
Screen Shot 2022-02-16 at 3 34 02 PM

So you can clearly see where this new expansion starts for large args compared to old versions.
Unfortunately now considering I just updated the medium argument branch for `sincos` it is now faster than the large arguments... (originally I started this PR with the large arguments being 2x faster.....
```julia
# large arguments (very accurate x > 75)
julia> @benchmark besselj0(x) setup=(x=rand()*1000)
 Range (min … max):   8.008 ns … 72.030 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     28.195 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   27.833 ns ±  2.382 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                    ▁▄ ▂             ▇█▃▂▁    ▁
  █▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▇████▄▅▁▁▃▁▁▁▃▁▃▁▁████████ █
  8.01 ns      Histogram: log(frequency) by time      30.9 ns <

# medium arguments (not as accurate)
julia> @benchmark besselj0(x) setup=(x=rand()*100)
 Range (min … max):   7.924 ns … 49.091 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     21.813 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   22.679 ns ±  4.562 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                         █                     
  ▄▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▅█▂█▂▂▂▂▂▂▁▂▂▂▂▂▂▁▁█▄ ▂
  7.92 ns         Histogram: frequency by time        28.6 ns <

```
Though they are both still much faster than SpecialFunctions.jl
```julia
julia> @benchmark SpecialFunctions.besselj0(x) setup=(x=rand()*1000)
 Range (min … max):   9.093 ns … 189.690 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     51.719 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   52.002 ns ±   3.241 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                       ▆█       
  ▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▃██▄▃▃▂▂ ▂
  9.09 ns         Histogram: frequency by time         56.4 ns <

```

Are we ok with these tradeoffs between speed and accuracy?? It might be good to set up strict rules to follow on this so any optimization must be accurate to x amount (relative or absolute error threshold)? Potentially maybe offer fast versions... unsure

src/besselj.jl
oscardssmith: There's a much better way to do this. Use
```
mycos(x, xn)
    n, r = Base.Math.rem_pio2_kernel(x)
    hi, lo = r.hi, r.lo
    n2, r = Base.Math.rem_pio2_kernel(hi + (r + lo))
    n = (n+n2)&3
    if n == 0
        return cos_kernel(y)
    elseif n == 1
        return -sin_kernel(y)
    elseif n == 2
        return -cos_kernel(y)
    else
        return sin_kernel(y)
    end
end
```

heltonmc: I'm getting several errors running this. Is xn supposed to be y?
Also    ` n2, r = Base.Math.rem_pio2_kernel(hi + (r + lo))` return a method error for me (r is some `Base.Math.DoubleFloat64`)

oscardssmith: oops, fixed.

heltonmc: Ok works now but gives worse errors...
Screen Shot 2022-02-16 at 4 00 13 PM

heltonmc:
```julia
julia> cos(big"100.0" + big"-0.10")
0.8074586576995466543840048046567971916342575898589744635569386441759010692980412

julia> mycos(100.0, -0.10)
0.8074586576995554

julia> cos(100.0 + (-0.10))
0.80745865769955
```

oscardssmith: Here's a version that gets within about 1 ulp (where tested, there's probably a case where it does slightly worse).
```
function mycos(x, xn)
    s = x + xn
    n, r = Base.Math.rem_pio2_kernel(s)
    lo = r.lo - ((s - x) - xn)
    hi = r.hi + lo
    y = Base.Math.DoubleFloat64(hi, r.hi-hi+lo)
    n = n&3
    if n == 0
        return Base.Math.cos_kernel(y)
    elseif n == 1
        return -Base.Math.sin_kernel(y)
    elseif n == 2
        return -Base.Math.cos_kernel(y)
    else
        return Base.Math.sin_kernel(y)
    end
end
```

oscardssmith: Some analysis suggests that this should be extremely accurate as long as `abs(x+xn) < 10^14`

heltonmc: Yes!! This looks great. Awesome stuff and is really fast.. This is now 2.8x faster than SpecialFunctions.jl and really accurate!
Screen Shot 2022-02-17 at 12 14 41 AM

src/misc.jl
```diff
@@ -0,0 +1,39 @@
+# function to more accurately compute cos(x + xn)
+# when x >> xn
```
oscardssmith, line 2: this doesn't require x>>xn

oscardssmith: also, you should probably not name it `mycos`, maybe `cos_sum`?

heltonmc: Sounds good! Meant more that cos(x + xn) is inaccurate when x >> xn not `mycos`. I will rename and remove that.

oscardssmith: the problem with `cos(x+xn)` isn't really that `x>>xn`, but that `abs(x+xn) >> pi/2`.

oscardssmith:
Is this ready to merge?

oscardssmith:
I just realized that this introduces type instability for Float32 (and uses too many terms).

src/besselj.jl
```diff
@@ -1,42 +1,77 @@
...
+    else
+        xinv = inv(x)
+        x2 = xinv*xinv
+
+        p = (one(T), -1/16, 53/512, -4447/8192, 5066403/524288)
+        p = evalpoly(x2, p)
+        a = SQ2OPI(T) * sqrt(xinv) * p
+
+        q = (-1/8, 25/384, -1073/5120, 375733/229376, -55384775/2359296)
+        xn = muladd(xinv, evalpoly(x2, q), - PIO4(T))
```
oscardssmith, line 64: This subtraction of pi/4 would be better handled in `cos_sum` since we could just subtract 1 from `n` to shift the quadrant by 1.

heltonmc: Yes definitely better to do it that way!

heltonmc:
Let me look. We don't use that for the `Float32` right now but I think those versions could be made more accurate anyway. It's probably best to use this new way to calculate the `Float32` using less terms but computed in Float64 anyway. I would imagine these asymptotic expansions can cover a significantly wider range using `Float32` precision but need to test when that cutoff is first.

PR #14, opened 2022-02-25
besselj0 and friends for medium arguments 

heltonmc:
Continuing from #13. These expansions work well into the Fresnel region (x > order) though this isn't particularly true for low orders ( nu = 0, 1). We need to handle these with care. The actual cutoff range is about x > 25.0 for these to work.

The challenge is we need a few more coefficients... mainly we need to compute 
```julia
alpha = atan(-Q0/P0)
beta = sqrt(P0^2 + Q0^2)
# where Q0 and P0 are..
p = 1 - 9/128x^2 + 3675/32768x^4 - 2401245/4194304x^6 + 13043905875/2147483648x^8 - 30241281245175/274877906944x^10 + 213786613951685775/70368744177664x^12 - 1070401384414690453125/9007199254740992x^14 + 57673297952355815927071875/9223372036854775808x^16
q = -1/8x + 75/1024x^3 - 59535/262144x^5 + 57972915/33554432x^7 - 418854310875/17179869184x^9 + 1212400457192925/2199023255552x^11 - 10278202593831046875/562949953421312x^13 + 60013837619516978071875/72057594037927936x^15 - 3694483615889146090857721875/73786976294838206464x^17
```
I was able to finally (after some tedious calculations) express `beta` in terms of a single polynomial, therefore `beta` can be computed simply with a single `evalpoly`. Now `alpha` should also be able to do this.... but much trickier. We have to divide the power series and then expand `atan`. Or as discussed should be able to do a minimax here as discussed offline. 

Unfortunately, I have not been able to get good results for this yet doing a polynomial approximation....

However, as is the results are significantly better.
Screen Shot 2022-02-25 at 1 23 39 PM
I also caught an error in the coefficients that the original paper gave. Updating this improves the accuracy as well.


heltonmc:
Unfortunately, the `atan` and having two polynomials to calculate slows this calculation down. We should be able to remove this though... I'll keep working on this.

oscardssmith:
Do you have bounds on `p/q`? If so, you might be able to beat `atan` by a bunch.

heltonmc:
Ok so if
```julia
p = 1 - 9/128x^2 + 3675/32768x^4 - 2401245/4194304x^6 + 13043905875/2147483648x^8 - 30241281245175/274877906944x^10 + 213786613951685775/70368744177664x^12 - 1070401384414690453125/9007199254740992x^14 + 57673297952355815927071875/9223372036854775808x^16
q = -1/8x + 75/1024x^3 - 59535/262144x^5 + 57972915/33554432x^7 - 418854310875/17179869184x^9 + 1212400457192925/2199023255552x^11 - 10278202593831046875/562949953421312x^13 + 60013837619516978071875/72057594037927936x^15 - 3694483615889146090857721875/73786976294838206464x^17
```
We want a good polynomial estimate for something like `alpha = atan(-q/p)` for `25.0 < x < 75.0`. In that case, `0.004995896094423783 > -q/p > 0.00166651397665257`.

We do know that the first few terms are `alpha = 1/8x  - 25/384x^3 + 1073/5120x^5 - 375733/229376x^7 + 55384775/2359296x^9` Just trying to work out two or three more terms for this

oscardssmith:
Here you go,
```
using Remez
p = (1, -big(9)/128, big(3675)/32768, -big(2401245)/4194304, big(13043905875)/2147483648, -big(30241281245175)/274877906944, big(213786613951685775)/70368744177664, big(-1070401384414690453125)/9007199254740992, big(57673297952355815927071875)/9223372036854775808)
q = (-1/8, big(75)/1024, -big(59535)/262144, big(57972915)/33554432, -big(418854310875)/17179869184, big(1212400457192925)/2199023255552, -big(10278202593831046875)/562949953421312, big(60013837619516978071875)/72057594037927936, -big(3694483615889146090857721875)/73786976294838206464)
f(x) = big(pi)/2-atan(-x*evalpoly(x*x, q)/evalpoly(x*x,p))
ratfn_minimax(x->f(sqrt(inv(x)))/sqrt(inv(x)), (inv(75)^2,inv(25)^2), 7, 0, (x,y)->sqrt(inv(x)))[1]
```
which gives
```
(0.1248852157943067033976124885207544013338361136050851240671434330360917760969508, -0.0009453699196465613464051070957028073864060688256309281305950690263708416153584159, 2.090652070386382568326123434103603187495192672139854437682469629722064389437547e-05, -8.542107797174051754942427187234643328955569618114482086921902361225973565375501e-07, 5.848596407775187490378388480839851353591119668460879337227688694400032613957985e-08, -6.909155983613888396147312686058543769738730774991659856401364356635301505635752e-09, 1.643173083545566687940485529708311841457038889858748960174833513199870876664362e-09)
```
as coefficients.

oscardssmith:
How are the minimax coefficients compared to rational ones?

heltonmc:
Alright this is now ready to go. It is highly accurate and have removed all arcan calls so it is very fast.

```julia
julia> @benchmark besselj0(x) setup=(x=rand()*1000)
 Range (min … max):   9.009 ns … 71.321 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     17.810 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   18.226 ns ±  1.810 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁                                  ▄█▇▇ ▁▂ ▂          ▄▃▄▁▂ ▂
  █▃▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█████████▆▅▇▆▇▆▃▃▃▄█████ █
  9.01 ns      Histogram: log(frequency) by time      22.8 ns <

julia> @benchmark SpecialFunctions.besselj0(x) setup=(x=rand()*100)
 Range (min … max):   5.958 ns … 104.541 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     50.042 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   49.345 ns ±   6.502 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂                                                   ▃▂█▆▃▃▂  ▁
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁████████ █
  5.96 ns       Histogram: log(frequency) by time      54.2 ns <

```

Now, there is that other piece. Essentially if we want to include the subtraction within the `cos_sum`.......
```julia
xn = muladd(xinv, evalpoly(x2, q), - PIO4(T))
b = cos_sum(x, xn)
```

heltonmc:
I'm still trying to get the minimax coefficients to work correctly for this...... for some reason was having trouble getting them to work. Though I was finally able to get the power series for arctan figured out so the coefficients used now are exact.

oscardssmith:
Do the coefficients I gave not work?

heltonmc:
O yes the polynomials were in inverse x. Were you also trying to fit with the subtraction piece ` muladd(xinv, evalpoly(x2, q), - PIO4(T))`? I missed that looking over the code again. I'm not for sure how much it matters now since we have the exact coefficients to use unless we want to try to reduce order or include the pi/4

oscardssmith:
Can you give me 2 more exact coefficients? That should be enough for me to search for a good minmax one. I think that using a minmax polynomial will probably give you 1 or 2 fewer terms. Including the `pi/4` probably won't help much.

heltonmc:
So `tan(alpha) = -(-1/8, 25/384, -1073/5120, 375733/229376, -55384775/2359296, 24713030909/46137344, -7780757249041/436207616)` are the exact coefficients for this. Though, with some quick benchmarks the performance is only about 1-5% faster with 1 or 2 less terms. Not sure how much there is to gain on this as the `cos_sum` is 55% of the runtime right now.

oscardssmith:
Makes sense (and I don't think there's anything to gain on `cos_sum`).

heltonmc:
Ok I've updated benchmarks. This results in a pretty substantial speed increase. 1.8x -> 2.8x faster than SpecialFunctions while decreasing the error significantly. 

Ready to merge?

oscardssmith:
I think it is, that said, last night I was looking a little at possible gains to `cos_sum` and one thing that can be done is using
```
@inline function reduce_pi02_med(x::Float64)
    pio2_1 = 1.57079632673412561417e+00

    fn = round(x*(2/pi))
    r  = muladd(-fn, pio2_1, x)
    w  = fn * 6.07710050650619224932e-11
    y = r-w
    return unsafe_trunc(Int, fn), DoubleFloat64(y, (r-y)-w)
end
```
instead of `Base.Math.rem_pio2_kernel`, which is valid for `10<x<2^20` and is a decent amount faster.

heltonmc:
Dropping that in instead is about 5% faster for `besselj0` which changes comparisons to SpecialFunctions slightly. 2.9x->3.1x faster.

PR #15, opened 2022-02-28
faster besselj0 for x<25

oscardssmith:
this is about 2x faster on my computer where applicable. This also should be done for `besselj1` and `bessely0`, `bessely1`


oscardssmith:
Here's a plot of relative errors.
image

heltonmc:
Though it's another branch. It might be best to keep the power series for very small arguments to ensure that `besselj0(0.0)= 1.0` (test that fails currently).

oscardssmith:
Fixed.  I used 1 to few terms, and didn't force the 1st coefficient to be 1.0.

oscardssmith:
Also, for future reference, here is the code used to generate the polynomials
```
using ArbNumerics
function r(x,root)
    abs(x)<eps(Float64) && return -besselj(1,ArbFloat(x)+root[1]+root[2])
    return besselj(0,ArbFloat(x)+root[1]+root[2])/x
end
function r2(x,root)
    return besselj(0,ArbFloat(x)+root[1]+root[2])
end
polys = []
for i in 2:2:16
    poly = Float64.(Tuple(ratfn_minimax(x->r(x,pts[i]), ((pts[i-1][1]-pts[i][1]-.1)/2, (pts[i+1][1]-pts[i][1]+.1)/2), 12, 0)[1]))
    push!(polys, (0.0, poly...))
    if i!=16
        poly = Float64.(Tuple(ratfn_minimax(x->r2(x,pts[i+1]), ((pts[i][1]-pts[i+1][1]-.1)/2, (pts[i+2][1]-pts[i+1][1]+.1)/2), 13, 0)[1]))
        push!(polys, poly)
    end
end
Tuple(polys)
```
for the 0-pi/2 range, the invocation was `(1.0, Float64.(Tuple(ratfn_minimax(x->(besselj(0,sqrt(x))-1)/x, (1e-20,nextfloat(pi/2)^2), 7, 0)[1]))...)`

heltonmc:
We may also want to extend this range just a bit more to blend with the asymptotic expansion better. I had chosen that cutoff to blend well with the previous method but we may want to use this for x < 26.0 now ? 

Screen Shot 2022-02-27 at 10 59 49 PM

oscardssmith:
Makes sense. Do you want to do the honors, or should I?

heltonmc:
If you have it set up that might be easiest! Otherwise, I can look more at this tomorrow and also for `J1`, `Y0`, and `Y1`. But this is awesome. Hoping to finally finish these special cases! (at least for Float64)

oscardssmith:
updated (the post with computation details and error plots are also updated).

oscardssmith:
The last 2 points are wrong. They should be
`(25.903672087618382, 4.894530726419825e-16)` and `(26.1,0.0)`. (you want the last point to be the reflection of the last root around your cutoff value).

heltonmc:
It looks like I also misspoke were `pts` contains the roots of besselj0 and besselj1? Unsure what the first point is though.

oscardssmith:
The derivative of `besselj0` is `-besselj1`, so extrema of `besselj0` are roots of `besselj1`. The first point is a reflection of the second point around `x=pi/2` (which is the lower cutoff)

oscardssmith:
This now is implemented for `besselj1` although something with my polynomials is slightly broken, which is causing spikes in error at the extrema of the function which really shouldn't be happening.

oscardssmith:
This is now failing tests because SpecialFunctions is too inaccurate near `3.8317059702075125`. The tests probably need to use `Arb`.

heltonmc:
I agree that we should probably switch the tests over to Arb for all this. I actually had a draft of this early on but ran into https://github.com/JeffreySarnoff/ArbNumerics.jl/issues/48 causing some issues. Now that it is fixed we should move over.
Though, for this test `besselj1` is computed in higher precision I believe using SpecialFunctions.jl as they provide `BigFloat` implementations. I'll have to check this some more.

src/besselj.jl
```diff
@@ -115,33 +218,28 @@ function besselj1(x::Float64)
     isinf(x) && return zero(x)
 
     if x <= 5.0
```
heltonmc, line 220: This branch should be less than ~26? Or wherever the minimax was done up to.

oscardssmith: good catch.

src/constants.jl
```diff
@@ -59,51 +40,6 @@ const YQ_y1(::Type{Float64}) = (
...
-    4.52228297998194034323E11, -8.99971225705559398224E8
-)
-const RQ_j1(::Type{Float64}) = (
-    5.32278620332680085395E18, 8.95222336184627338078E16,
-    7.84369607876235854894E14, 4.74914122079991414898E12,
-    2.21511595479792499675E10, 8.35146791431949253037E7,
-    2.56987256757748830383E5, 6.20836478118054335476E2,
-    1.00000000000000000000E0
-)
-const PP_j1(::Type{Float64}) = (
```
heltonmc, line 73: Unfortunately several of these constants are used for the `bessely` implementations.

oscardssmith: it will work out once we replace those implementations as well :)

heltonmc: O most definitely!! Didn't know if you wanted to do that in separate PR or all at once 😃

src/bessely.jl
```diff
@@ -63,19 +63,22 @@ function _bessely0_compute(x::Float64)
...
+            q1 = (-1/8, 25/384, -1073/5120, 375733/229376, -55384775/2359296, 24713030909/46137344, -7780757249041/436207616)
+            p = evalpoly(x2, p1)
+            q = evalpoly(x2, q1)
         else
-            p = (one(T), -1/16, 53/512, -4447/8192)
-            q = (-1/8, 25/384, -1073/5120, 375733/229376)
+            p2 = (one(T), -1/16, 53/512, -4447/8192)
+            q2 = (-1/8, 25/384, -1073/5120, 375733/229376)
+            p = evalpoly(x2, p2)
+            evalpoly(x2, q2)
```
heltonmc, line 77: q = evalpoly(x2, q2) hopefully fixes that test

src/besselj.jl
```diff
@@ -3,32 +3,98 @@
...
+        (14.930917708487787, -7.070514505983074e-16),
+        (16.470630050877634, -1.619019544798128e-15),
+        (18.071063967910924, -9.658048089426209e-16),
+        (19.615858510468243, -1.004445634526616e-15),
+        (21.211636629879260, 4.947077428784068e-16),
+        (22.760084380592772, -4.925749373614922e-16),
+        (24.352471530749302, 9.169067133951066e-16),
+        (25.903672087618382, 4.894530726419825e-16)
+)
+@inline J0_POLYS(::Type{Float64}) = (
```
heltonmc, line 44: Is there a reason we sometimes inline these, sometimes we don't, and sometimes we use constants for the polynomials? It might be good to have a defined way going forward that is consistent. Though this one is slightly different as we have to determine which polynomial at runtime. Does this affect compile time?

heltonmc: Also, I would be a fan of moving the polynomials to their own file and slightly different file structure to make the core code easier to read. I think this approach could also be used for `Double64` implementations that would also take up a lot of space in main file.

src/besselj.jl
```diff
@@ -114,34 +217,29 @@ function besselj1(x::Float64)
     x = abs(x)
     isinf(x) && return zero(x)
```
heltonmc, line 218: We should probably be consistent with the structure between `besselj0` and `besselj1`. Either have the isinf check in the large argument branch (where I check if xinv is zero) or just have it at beginning.

src/besselj.jl
```diff
@@ -114,34 +217,29 @@ function besselj1(x::Float64)
...
-        w = 5.0 / x
-        z = w * w
-        p = evalpoly(z, PP_j1(T)) / evalpoly(z, PQ_j1(T))
-        q = evalpoly(z, QP_j1(T)) / evalpoly(z, QQ_j1(T))
-        xn = x - THPIO4(T)
-        sc = sincos(xn)
-        p = p * sc[2] - w * q * sc[1]
-        return p * SQ2OPI(T) / sqrt(x)
+    if x <= 26.0
+        x <= pi/2 && return x*evalpoly(x*x, J1_POLY_PIO2(T))
```
heltonmc, line 221: Also, do we want to set a code style for this repo? I think in general I've been trying to use blue but sometimes I obsess over spaces too much 😅

heltonmc: I usually _try_ to follow https://github.com/jrevels/YASGuide. I can make some spacing changes if that's ok.

Otherwise I think this is ready to merge?

oscardssmith: sounds good. (make any changes you want).

heltonmc:
I added a fix to the sign error for `besselj1` and updated spacing, moved polys to only file. The sign error could be done in a different way perhaps. This is ready to merge on my end.

src/besselj.jl
```diff
@@ -160,72 +88,26 @@ function besselj0(x::Float32)
...
+        x <= pi/2 && return x * evalpoly(x * x, J1_POLY_PIO2(T))
+        n = unsafe_trunc(Int, TWOOPI(T) * x)
         root = @inbounds J1_ROOTS(T)[n]
         r = x - root[1] - root[2]
-        return evalpoly(r, @inbounds J1_POLYS(T)[n])
+        return evalpoly(r, @inbounds J1_POLYS(T)[n]) * s
     else
         xinv = inv(x)
-        x2 = xinv*xinv
+        iszero(xinv) && return zero(T)
```
oscardssmith, line 109: technically this should probably return `xinv` so it returns `-0.0` when applicable.

heltonmc: I believe since we have `x=abs(x)` this would still always return `0.0` so might need to handle that separately. I'm actually starting to not appreciate that syntax of overwriting that variable and prefering `xabs = abs(x)`. Thoughts?

oscardssmith: I don't think it really matters. I would probably just merge this as is.

Issue #18, opened 2022-05-10
Anything to take from Fortran?


oscardssmith:
https://github.com/arjenmarkus/specfunc-fullerton

heltonmc:
Looks like they only give routines for the modified Bessel function's of integer order. 

`besselk` is done with forward recurrence similar to us but they don't use the uniform asymptotic expansion for large orders so that routine will be much slower than ours at large orders.

`besseli` is done with backward recurrence which on quick glance it looks like they have to come back all the way to `besseli0` which will also be very slow for large argument.

I think our routines would be better but I haven't run their code to confirm obviously.

oscardssmith:
makes sense. That's kind of what I had expected, but figured it was worth a look.

heltonmc:
A general fall back method though for the modified versions would be helpful for non-integer orders.
I spent some time today looking at the details in https://arxiv.org/pdf/1209.1547.pdf for numerical integration that will work for any order.... the challenge I found was a very significant number of terms were needed (n>150) to get good accuracy which made it quite slow.

oscardssmith:
is it better for `Float16` or `Float32` arguments?

PR #19, opened 2022-05-11
Float16 version of besselj0

heltonmc:
From the discussion in #18, it might be useful to also provide lower precision `Float16` as well. Here is a first pass at this for `besselj0` it has a maximum relative error of 1.303792e-5 which is good enough for `Float16`.

It is also much faster...
```julia
julia> @benchmark Bessels.besselj0_Float16(x) setup=(x=rand()*10)
 Range (min … max):   3.618 ns … 31.208 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     11.285 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):    9.753 ns ±  5.091 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █               ▁▁  ▆▆▄▅    ▆ ▅   ▃                    ▂    ▂
  █▃▁▃▅▄▃▄▁▁▁▁▁▁▁▁██▁▁████▅█▆▄█▄█▁▇▅█▇▇██▇▇▆▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▄█ █
  3.62 ns      Histogram: log(frequency) by time      25.2 ns <

julia> @benchmark Bessels.besselj0(x) setup=(x=rand()*10)
 Range (min … max):   4.914 ns … 34.692 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     25.608 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   22.534 ns ±  7.523 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                        █      
  ▅▂▁▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▂▅▂▂ ▂
  4.91 ns         Histogram: frequency by time        27.2 ns <

```
The function right now was doing everything in `Float64` so need to go through and fix this. It is probably best to do all these calculations in `Float32`? I'm not for sure if there is that much of a performance gain from doing this in `Float16`. I think the power series for small arguments can be done in low precision but will need to be careful again with the cos(x+xn) term.

And a side note. I tried the integral version of besselj0 = cos(-xsin(\theta)) d\theta integrated from 0 to pi and found that for small arguments this could be evaluated fairly quickly but large arguments it oscillates too much to be worth it.


oscardssmith:
Using Float64 for the math is reasonable, it should be similar speed (at least for scalar stuff)

heltonmc:
Sounds good! I think I'll plan to do all the calculations in `Float32` as that should be plenty for this precision? Might be helpful for GPU computing? Though, maybe I should spend some time on the pure 'Float32' implementation instead which would only require two more digits of accuracy.... I'll play around with that as simply converting the Float16 to Float32 and converting it back takes a non-negligible amount of time here it appears...

PR #20, opened 2022-05-12
Implementation for besselj of any (positive) real order and argument

heltonmc:
Here is an implementation for `besselj` and `bessely` for large arguments based on https://www.math.ucdavis.edu/~bremer/papers/bessel8.pdf. Technically, this approach works for any argument in the Fresnel regime (x>nu) though based on the results in the paper it is quite slowly converging when x<2nu.

This is just a draft as getting these coefficients right was tricky but they can produce very accurate results depending on the argument. 
```julia
julia> dnu = big"4"
4
julia> x = big"20.0"
20.0
julia> 1 -  besselj(dnu, x) / besselj_large_argument(dnu, x)
-3.071266864290228855849154260328918009213010091795647051576856584355915252396212e-16
julia> x = big"200.0"
200.0

julia> 1 -  besselj(dnu, x) / besselj_large_argument(dnu, x)
-3.65422644136090656921668723356207962364326830807318559819150864916674755951513e-36
```

The advantage is this will work for integer and non-integer orders and complex orders. It will also produce both `bessely` and `besselj` in essentially the same time. 

I plan to build these implementations piecewise and fit them together. There will have to be several algorithms to cover all the ranges and testing. I have not yet optimized this approach as I'm just focusing on getting this working for now.

Edit: and for `bessely` we just return `sin` isntead of `cos`


heltonmc:
We'll have to scale the values we don't want to compute `mu^8` here but probably `(mu/x)^8` to avoid overflow. The key for these asymptotic expansions is to avoid loss of precision when using Float64. In practice we will have to balance when we want to employ this approach. Like for when x>5nu or something larger. The larger x is than nu the less terms (s0, s1, s2...) you will need. Probably 4 is sufficient if x>15nu but will have to balance this depending on how fast the other methods are for when  x~nu

heltonmc:
I've added two more coefficients s11 and s12. This improved the accuracy of previous examples to 
```julia
julia> x = big"20.0"
20.0
julia> 1 -  besselj(dnu, x) / besselj_large_argument(dnu, x)
-1.617685474657480344179524558783495937072544030980921933922201321722990788375528e-17

julia> x = big"200.0"
200.0

julia> 1 -  besselj(dnu, x) / besselj_large_argument(dnu, x)
-2.726926944584588714318390650413623099583163111074598679578561149580182018603696e-41
```
This does improve the accuracy, however two more coefficients only gave 1-2 more digits of accuracy for smaller `x` whereas it gave over 5 digits more of accuracy for large x. This suggests that although the method works for x>nu it requires a lot of terms for small x and will probably want to employ this only when x is some multiple of nu (x>5nu). We will want to choose this cutoff well and compare the speed to other methods when x~nu.... Though, for x>>nu this is a viable solution for quadruple precision.

src/besselj.jl
```diff
@@ -176,3 +176,16 @@ function besselj_large_argument(v, x::T) where T
     b = sqrt(2 / T(pi)) / sqrt(αp * x)
     return cos(α)*b
 end
+
+function besselj_small_arguments_orders(v, x::T) where T
+    MaxIter = 100
+    out = zero(T)
+    a = inv(factorial(v)) * (x/2)^v
```
oscardssmith, line 183: `(x/2)^v/factorial(v)`?

heltonmc:
I'm going to also slow add the small argument approximation using the power series. This has some current roundoff error that needs to be dealt with...
```julia
julia> besselj_small_arguments_orders(20.0, 0.1)
3.919437720858622e-45
julia> Float64(besselj_small_arguments_orders(big"20.0", big"0.1"))
3.9194377208586175e-45
```
But will converge quickly for small values of x. I will need another version using the logarithms for large orders though.
There is a bit of a hiccup in performance though.
```julia
julia> @btime besselj_small_arguments_orders(20, 0.1)
  42.634 ns (0 allocations: 0 bytes)
3.919437720858622e-45

julia> @btime besselj_small_arguments_orders(20.0, 0.1)
  276.144 ns (0 allocations: 0 bytes)
3.919437720858622e-45
```
The issue seems to be from the factorial function. I'm assuming if nu is an integer it uses a lookup table and if it's a float it doesn't...
```julia
julia> @btime factorial(20)
  2.125 ns (0 allocations: 0 bytes)
2432902008176640000

julia> @btime factorial(20.0)
  211.650 ns (0 allocations: 0 bytes)
2.43290200817664e18
```

oscardssmith:
Also, does this method require an exact `gamma` computation? If it is iterative and will recover the error, you could get a very fast rough approximation (especially if you have bounds on `v`).

heltonmc:
> `factorial` in Julia is only defined for `::Integer`. Do you want `gamma`?

Yes - I originally had `gamma(v + one(T))` there and then I can use the loggamma version. Was avoiding the SpecialFunctions dependency but should just go ahead using gamma here because we use it in several functions

heltonmc:
I'm not exact sure on the accuracy yet but we will definitely need a much faster version as gamma takes the majority of the runtime right now. This routine can produce very accurate results so I'm assuming we'll need a fairly accurate version. 
```julia
julia> @benchmark besselj_small_arguments_orders(100.0, 1.5)
BenchmarkTools.Trial: 10000 samples with 346 evaluations.
 Range (min … max):  260.237 ns … 603.081 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     262.523 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   263.805 ns ±   7.834 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂▆▆▇██▄▂         ▁▁▃▃▂▂▁                                      ▂
  █████████▆▆▆▆▆▆▆▇███████▇▆▆▆▇▇▇▆▆▅▅▅▄▆▅▆▆▅▆▅▅▅▅▄▄▄▁▃▁▄▁▃▅▃▃▄▄ █
  260 ns        Histogram: log(frequency) by time        291 ns <

julia> @benchmark gamma(100.0)
BenchmarkTools.Trial: 10000 samples with 535 evaluations.
 Range (min … max):  211.604 ns … 380.062 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     212.228 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   213.337 ns ±   4.939 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅█▆▄▆▅          ▂▃▁ ▂                                         ▂
  ██████▇▆▆▅▅▇▆▆▆▅███▇█▇▆▇▇▆▇█▇▆▅▆▄▄▄▅▄▄▅▃▃▄▄▄▄▄▅▄▅▃▄▅▃▄▅▅▁▄▃▄▃ █
  212 ns        Histogram: log(frequency) by time        231 ns <

```

oscardssmith:
```
function mygamma(x)
    if x<.5
        return pi/(sinpi(x)*mygamma(1-x))
    end
    if x<11
        # innaccurate 
        return mygamma(x+1)/x
    end
    x-=1
    if x<20.5
        poly = evalpoly(inv(x), (1., 0.08333333333650161, 0.003472221906979175, -0.002681313976042538, -0.00022977090009804665, 0.0007879538510356031, 4.1092075095252335e-5, -0.0004988424489412549))
    else
        poly = evalpoly(inv(x), (1.0,0.08333333333346536, 0.003472222180925872, -0.002681322445355217, -0.000229731127876239, 0.0007914216474948752, -3.547263379392065e-5))
    end
    s2pi = sqrt(2*pi*x)
    logxhi,logxlo = Base.Math._log_ext(reinterpret(UInt64, x))
    xyhi, xylo = Base.Math.two_mul(logxhi-1,x)
    xylo = muladd(logxlo, x, xylo)
    hi = xyhi+xylo
    exp_term = Base.Math.exp_impl(hi, xylo-(hi-xyhi), Val(:ℯ))
    return s2pi*exp_term*poly
end
```
This is within 3 ULP for x>11 (I need a good branch for x<11, and runs in 12ns.

src/besselj.jl
```diff
@@ -150,3 +150,29 @@ function besselj1(x::Float32)
...
+    s9 = (-715μ^9 + 3026595μ^8 - 2163210764μ^7 + 597489288988μ^6 - 79979851361130μ^5 + 5536596631240042μ^4 - 194026764558396188μ^3 + 3095397360215483916μ^2 - 17285630127691126691μ + 14378802319925055947) / 17179869184
+    s10 = (-2431μ^10 + 14318590μ^9 - 14587179035μ^8 + 5925778483368μ^7 - 1216961874423502μ^6 + 137164402798287604μ^5 - 8556293060689145118μ^4 + 281576004385356401192μ^3 - 4334421752432088249971μ^2 + 23801928703130666089534μ - 19740662615375374580231) / 274877906944
+    s11 = (-4199μ^11 + 33302269μ^10 - 46575001473μ^9 + 26616800349043μ^8 - 7938052610367302μ^7 + 1355921534379626242μ^6 - 136097258378143928354μ^5 + 7882145023568373727078μ^4 - 247560495216085669748963μ^3 + 3708133715826433118900337μ^2 - 20097626064642945009122253μ + 16629305448257355302267575) / 2199023255552
+    s12 = (-29393μ^12 + 305569628μ^11 - 569135407698μ^10 + 441734362333708μ^9 - 183408097342762543μ^8 + 45038004674954790968μ^7 - 6784356890740220944060μ^6 + 626292216963766324300664μ^5 - 34303895772550410316410943μ^4 + 1039206680182498322210164588μ^3 - 15233583205322702897380357650μ^2 + 81699279492428006742420030332μ - 67471218624230362526181277601) / 70368744177664
```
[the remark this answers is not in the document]
heltonmc: I guess there is no way to better combine all these calculations than doing evalpoly for each `s1, s2...`. I'll update this

heltonmc:
Quick question. I am getting an error for `Base.Math._log_ext`...
```julia
ERROR: MethodError: no method matching _log_ext(::UInt64)
Closest candidates are:
  _log_ext(::Float64) at ...
```
Is this defined on later versions or am I just making mistake?

src/besselj.jl
```diff
@@ -151,41 +151,72 @@ function besselj1(x::Float32)
...
+        s11 = evalpoly(μ, (16629305448257355302267575, -20097626064642945009122253, 3708133715826433118900337, -247560495216085669748963, 7882145023568373727078, -136097258378143928354, 1355921534379626242, -7938052610367302, 26616800349043, -46575001473, 33302269, -4199)) / 2199023255552
+        s12 = evalpoly(μ, (-67471218624230362526181277601, 81699279492428006742420030332, -15233583205322702897380357650, 1039206680182498322210164588, -34303895772550410316410943, 626292216963766324300664, -6784356890740220944060, 45038004674954790968, -183408097342762543, 441734362333708, -569135407698, 305569628, -29393)) / 70368744177664
+        s13 = evalpoly(μ, (81121053093143918050102987600099,  -98383136980007985475571939027975, 18503851640550924693860888790250, -1284587641197372537980918233562, 43642612323123398148213493825, -832261743620394930038568309, 9606500048207046619311900, -69893489053616393418780, 325411350127146969525, -959952773579760385, 1717262634951770, -1676314996330, 692939975, -52003)) / 562949953421312
```
oscardssmith, line 171: these coeffs are being stored as `Int128` (they're big) which won't be good for performance. You probably want to fix that.

src/besselj.jl
```diff
@@ -220,3 +220,16 @@ function log_besselj_small_arguments_orders(v, x::T) where T
...
     return logout
 end
+
+# valid when x < v (uniform asymptotic expansions)
+function besselj_debye(v, x)
+    T = eltype(x)
+    S = promote_type(T, Float64)
+    x = S(x)
+    n = v*log(v/x + sqrt((v/x)^2 - 1)) - sqrt(v^2 - x^2)
+    coef = inv(sqrt(2*T(pi))) * exp(-n) / (v^2 - x^2)^(1/4)
```
oscardssmith, line 230: `(v^2 - x^2)^(1/4)` will be much slower than `sqrt(sqrt(v^2 - x^2))` (especially since you already are using `sqrt(v^2 - x^2)`)

src/besselj.jl
```diff
@@ -220,3 +220,16 @@ function log_besselj_small_arguments_orders(v, x::T) where T
...
+    p = v / sqrt(v^2 - x^2)
+    p2  = v^2/fma(max(v,x), max(v,x), -min(v,x)^2)
```
oscardssmith, line 232: if `v<x`, isn't this just `v^2/(x^2 -v^2)`

oscardssmith:
I changed `_log_ext` in 1.9, so this is only valid on a somewhat recent version of master. If you want this to work for generic Julia versions, you would need to copy the implementation probably.

src/besselj.jl
heltonmc: Yes - but `x<v` so `v^2/fma(v,v, -x^2)`. Fixed

src/besselj.jl
heltonmc: Fixed now. I need to go through next and see if we are losing any precision anywhere which I have not done. Just getting these to 'work' for now in BigFloat.

src/besselj.jl
oscardssmith: You should make `fma(v,v, -x^2)` a variable and use it since it will be cleaner and guarantee that you don't compute the same thing multiple times.

heltonmc:
For the gamma function for smaller arguments we should probably employ range reductions or increases for that range. Do you have a good polynomial approximation say for the interval (1,2)? If not here is a version that uses an exact polynomial approximation for x>5 and the increases the range of smaller arguments to compute over the whole range.
```julia
function g(x)
    p = (1.0, 0.08333333333333333333333333333333333, 0.00069444444444444444444444444444444, 0.00065861992945326278659611992945326, −0.00053287817827748383303938859494415, 0.00079278588700608376534302460228386, −0.00184758189322033028400606295961969, 0.00625067824784941846328836824623616, −0.02901710246301150993444701506844402, 0.17718457242491308890302832366796470, −1.37747681703993534399676348903067470, 13.286928409862371, -155.73220540533757, 2180.0409025624585, -35926.11177910434)#, 688464.6155042856)#, -1.5180622383127486e7, 3.8163135467193574e8)
    c = 1.0
    while x < 5.0
        c /= x
        x += 1.0
    end

    xinv = inv(x)^2
    q = evalpoly(xinv, p)
    a = (x/Base.MathConstants.e * q)^x * sqrt(2*pi/x)
    return a*c
end
```
I believe we can improve this substantially. It has a max ULP of 9 from 0 to 11 but only because I'm losing precision. When done in BigFloat it has a max ULP of 0.09 in that window. I think we should combine these approaches? And then honestly this method should be used in SpecialFunctions?

oscardssmith:
The hard part here is that compensating the arithmetic here will likely be fairly slow.

heltonmc:
Sounds good! For what it's worth I also tried using numerical integrals on Talbot contours following "COMPUTING THE GAMMA FUNCTION USING CONTOUR INTEGRALS AND RATIONAL APPROXIMATIONS". I couldn't get it to be accurate or fast enough for our interest (need complex numbers and several evaluations)

oscardssmith:
that's unfortunate. I think integration might only be useful for arbitrary precision.

src/besselj.jl
```diff
@@ -150,6 +150,29 @@ function besselj1(x::Float32)
...
+    nu == 1 && return besselj1(x)
+    if x < 30.0
+        if nu > 60.0
+            return log_besselj_small_arguments_orders(nu, x)
+        else
+            return besselj_small_arguments_orders(nu, x)
+        end
+    elseif 2*x < nu
+        return besselj_debye(nu, x)
+    elseif x > 2*nu
```
oscardssmith, line 170: isn't this just an `else`?

src/besselj.jl
```diff
@@ -238,3 +264,71 @@ function besselj_debye(v, x)
...
+        7.840334842744753003862e-4, 6.989332260623193171870e-5, -5.950237554056330156018e-4, -2.363848809501759061727e-5,
+        7.147391378143610789273e-4
+    )
+    w = w * evalpoly(w, s) + one(T)
+    # lose precision on following block
+    y = exp((x)) 
+    if x > 143.01608
+        isinf(x) && return x
+        # avoid overflow
+        v = x^(0.5 * x - 0.25)
```
oscardssmith, line 300: I think that just doing this unconditionally is likely to be better for performance.

heltonmc: Sounds good! We lose some precision here so we might be worth doing a routine like yours above in the thread. Ideally, we could have a version that would work on the long term stable release.

src/besselj.jl
heltonmc: There's really going to be 4 algorithms that will come into play.
1. Asymptotic expansion for x > 5nu
2. Power series for small arguments
3. Debye expansion for x < nu
4. Fall back routine for when x~nu that is undetermined.

The else branch will be the 4th algorithm just haven't gotten to that yet.

heltonmc:
This now "works" for values of x and nu between 0 and 100. It now also supports non-integer orders. This is still very rough so will piece this better and am still doing this in big floats so probably going to lose precision elsewhere. SpecialFunctions.jl doesn't provide higher precision for noninteger arguments and they only give about relative errors of 1e-13 for non-integer orders so it is difficult to test against that library. Should probably switch to Arb for our testing at this point.

So again this is broken up in several branches. But for nu < 100
1. x < 20.0: power series. For large orders use log of power series.
2. x > 1.85*nu: use asymptotic phase expansion. 
3. 2.5*x< nu: use debye expansion (not a uniform expansion so need the 2.5 factor buffer)
4. x <= nu: Shift order up and debye expansion and then use downward recurrsion
5. x > nu: Shift order down and use asymptotic expansion and upward recurssion (forward recursion is stable here)

At this point, we have to figure out the best way to handle large orders. The good thing is that the debye expansion is more accurate here so for x < nu we can use that. Though these aren't uniform so the transition region x~nu is always going to be difficult. Numerical integration here is brutal for higher orders as the integrand oscillates more. For large arguments we can also use the asymptotic expansions though there will be a lot of numerical errors for very large orders (v > 10^6). 

For large orders, the recursive based techniques become much slower especially if they depend on some multiplier to get in the proper range for the asymptotic expansions to work. Though it will depend on how accurate the debye expansions will be for x < nu at very large nu. The large argument asymptotic expansions I don't think will be feasible here as they always depend on some multiplier of 1.75x to be accurate.

With that comes the uniform asymptotic expansions for large orders that have been recently presented https://www.sciencedirect.com/science/article/pii/S0010465511003729. The difficulty is they now depend on the airy functions which are pretty slow provided by SpecialFunctions.... so we will have to rewrite those functions as well. I plan to implement this. But will take some more time.....

src/besselj.jl
```diff
@@ -265,6 +295,17 @@ function besselj_debye(v, x)
     return coef * Uk_poly_Jn(p, v, p2, T)
 end
 
+function besselj_debye_large_order(v, x)
+    T = eltype(x)
+    α = asech(x/v)
+    coef = exp(v*(tanh(α) - α))
```
oscardssmith, line 301: if x is large, (greater than 10) `tanh(α) - α` can just be `muladd(2, exp(-2x), -x)`

heltonmc:
We also need a better way to select how many terms to use in `_α_αp_asymptotic` as well as the u-polynomials. Right now I was planning on just adding more branches but seems like a hack to do it that way. 

 I'm wondering too how far to push these approximations by increasing the number of terms... for recurrence we need two evaluations and then recurrence. Unsure if it would be more efficient just to derive more terms in the asymptotic expansions since it very slowly converging when x~nu... though depending on how well the uniform asymptotic expansions work that might be mute.

src/besselj.jl
```diff
@@ -150,3 +150,222 @@ function besselj1(x::Float32)
...
+        abs(a) < eps(T) * abs(out) && break
+        a *= -inv((v + i + one(T)) * (i + one(T))) * t2
+    end
+    return out
+end
+
+# this needs a better way to sum these as it produces large errors
+# use when v is large and x is small
+# need for bessely 
+function log_besselj_small_arguments_orders(v, x::T) where T
```
heltonmc, line 266: We need a better way to calculate these power series. There is a lot of error propagation in the loops because the first 10 terms or so actually grow and oscillate...
```julia
julia> Bessels.log_besselj_small_arguments_orders(big"7", big"17.1")
0.1794205811198151334600691013458264941056284393401507757350616668754102873422111

julia> Bessels.log_besselj_small_arguments_orders(7, 17.1)
0.17942058112702328
```
Is there perhaps a better way to compensate for this?

oscardssmith: Is there any possible simplification from unrolling the first couple terms?

heltonmc: Ya I'm thinking because I have such tight bounds on when this is used `x<20` it might be worth just completely unrolling this and just using evalpoly for the max number of terms..? I think that would definitely be more accurate and probably faster than checking for convergence at each iteration?

oscardssmith: yeah that will be a lot faster since you can then make it only require 1 division rather than 1 division per step.

heltonmc: Should probably set this up better because I am still losing a lot of precision.
```julia
function t(v, x)
x2 = (x/2)^2

b = muladd(34, v, 1156)
a = muladd(-x2, inv(b), 1)

b = x2 / muladd(33, v, 1089)
a = muladd(-a, b, 1)
b = x2 / muladd(32, v, 1024)
a = muladd(-a, b, 1)
b = x2 / muladd(31, v, 961)
a = muladd(-a, b, 1)
b = x2 / muladd(30, v, 900)
a = muladd(-a, b, 1)
b = x2 / muladd(29, v, 841)
a = muladd(-a, b, 1)
b = x2 / muladd(28, v, 784)
a = muladd(-a, b, 1)
b = x2 / muladd(27, v, 729)
a = muladd(-a, b, 1)
b = x2 / muladd(26, v, 676)
a = muladd(-a, b, 1)
b = x2 / muladd(25, v, 625)
a = muladd(-a, b, 1)
b = x2 / muladd(24, v, 576)
a = muladd(-a, b, 1)
b = x2 / muladd(23, v, 529)
a = muladd(-a, b, 1)
b = x2 / muladd(22, v, 484)
a = muladd(-a, b, 1)
b = x2 / muladd(21, v, 441)
a = muladd(-a, b, 1)
b = x2 / muladd(20, v, 400)
a = muladd(-a, b, 1)
b = x2 / muladd(19, v, 361)
a = muladd(-a, b, 1)
b = x2 / muladd(18, v, 324)
a = muladd(-a, b, 1)
b = x2 / muladd(17, v, 289)
a = muladd(-a, b, 1)
b = x2 / muladd(16, v, 256)
a = muladd(-a, b, 1)
b = x2 / muladd(15, v, 225)
a = muladd(-a, b, 1)
b = x2 / muladd(14, v, 196)
a = muladd(-a, b, 1)
b = x2 / muladd(13, v, 169)
a = muladd(-a, b, 1)
b = x2 / muladd(12, v, 144)
a = muladd(-a, b, 1)
b = x2 / muladd(11, v, 121)
a = muladd(-a, b, 1)
b = x2 / muladd(10, v, 100)
a = muladd(-a, b, 1)
b = x2 / muladd(9, v, 81)
a = muladd(-a, b, 1)
b = x2 / muladd(8, v, 64)
a = muladd(-a, b, 1)
b = x2 / muladd(7, v, 49)
a = muladd(-a, b, 1)
b = x2 / muladd(6, v, 36)
a = muladd(-a, b, 1)
b = x2 / muladd(5, v, 25)
a = muladd(-a, b, 1)
b = x2 / muladd(4, v, 16)
a = muladd(-a, b, 1)
b = x2 / muladd(3, v, 9)
a = muladd(-a, b, 1)
b = x2 / muladd(2, v, 4)
a = muladd(-a, b, 1)
b = x2 / muladd(1, v, 1)
a = muladd(-a, b, 1)
return a*(x/2)^v * 1 / (SpecialFunctions.gamma(v+1))
end

julia> v = 7
7

julia> x = 17.1
17.1

julia> 1 - SpecialFunctions.besselj(v, x) / t(v,x)
4.896916205865409e-11

julia> v = big"7"
7

julia> x = big"17.1"
17.10000000000000000000000000000000000000000000000000000000000000000000000000006

julia> 1 - SpecialFunctions.besselj(v, x) / t(v,x)
2.11919566953799423385499459990264416821865652645594220138087841875073985051455e-19
```

src/besselj.jl
```diff
@@ -150,3 +150,222 @@ function besselj1(x::Float32)
...
+end
+
+# this needs a better way to sum these as it produces large errors
+# only valid in non-oscillatory regime (v>1/2, 0<t<sqrt(v^2 - 0.25))
+# power series has premature underflow for large orders
+# gives wrong answers for x > 20.0 (might want to fix this)
+function besselj_small_arguments_orders(v, x::T) where T
+    MaxIter = 1000
+    out = zero(T)
+    a = (x/2)^v / gamma(v + one(T))
```
oscardssmith, line 253: if you make this line `exp(muladd(log(x/2), v, -loggamma(v+1))`, this will work for both small and large `v`.

heltonmc: I had originally chosen to break these up because the log version was sometimes less accurate.
```julia
## true value
julia> out = 2.515386282716736709635E-7
2.515386282716737e-7

julia> function besselj_small_arguments_orders(v, x::T) where T
           MaxIter = 1000
           out = zero(T)
           a = (x/2)^v / gamma(v + one(T))
           t2 = (x/2)^2
           for i in 0:MaxIter
               out += a
               abs(a) < eps(T) * abs(out) && break
               a *= -inv((v + i + one(T)) * (i + one(T))) * t2
           end
           return out
       end
besselj_small_arguments_orders (generic function with 1 method)

julia> 1 - out/besselj_small_arguments_orders(10.0, 2.0)
2.220446049250313e-16

julia> function besselj_small_arguments_orders(v, x::T) where T
           MaxIter = 1000
           out = zero(T)
           a = exp(muladd(log(x/2), v, -loggamma(v+1)))
           t2 = (x/2)^2
           for i in 0:MaxIter
               out += a
               abs(a) < eps(T) * abs(out) && break
               a *= -inv((v + i + one(T)) * (i + one(T))) * t2
           end
           return out
       end
besselj_small_arguments_orders (generic function with 1 method)

julia> 1 - out/besselj_small_arguments_orders(10.0, 2.0)
-1.9984014443252818e-15
```

src/besselj.jl
oscardssmith: if you can find a maximum number of arguments, there is a major speedup here. You can multiply the whole equation by `(n+1)(n+2)(n+3)...` before hand (and you can fairly easily compensate this multiplication), and then you only have 1 division to do.

heltonmc: For `x<10` we need 25 terms and for `x<20` we need 40 terms. 

That was my current cutoff for this method `x<20` as for larger x it was harder to get accurate.

oscardssmith: you can do way better than that if you have bounds on the order, right?

heltonmc: Most definitely. But it doesn't equally scale for very large orders. We need about half the terms for orders above 60 which decrease slightly for increasing order.

For `x=20.0, nu = 0.5` we need 40 terms.
For `x = 20.0, nu = 60.5` we need 20 terms.

oscardssmith: if my math is correct, for `v>45` and `x<20`, 10 terms should suffice.

oscardssmith: that seems too pesimistic to me. Isn't the condition `x^2*gamma(v)/gamma(v+k)<2^-52`?

oscardssmith: also, we can deal with the cases where `x>>nu` using `besselj_large_argument`

heltonmc: That looks close to me quickly looking... I got those numbers from just outputting the amount of each interations from the loop it took to converge. 
```julia

julia> x=20.0; v = 0.5; k = 40
40

julia> x^2*gamma(v)/gamma(v+k)
5.51286033593326e-45
```

heltonmc: And for reference. The hypergeometric functions don't provide better accuracy..
```julia
using HypergeometricFunctions, SpecialFunctions
function besselj_hyper(v, x)
       a = HypergeometricFunctions.drummond0F1(v + 1, -x^2/4)
       return (x/2)^v / gamma(v+1) * a
end

julia> besselj_hyper(7.0, 17.1)
0.17915365638067893

julia> Float64(besselj_hyper(big"7.0", big"17.1"))
0.1791536563807948

julia> Float64(SpecialFunctions.besselj(big"7.0", big"17.1"))
0.1791536563807948
```

heltonmc:
Just to summarize a bit where we currently are.

There are three main algorithms:
1. Power series for small x
2. Debye expansion for x<nu
3. asymptotic expansion for x>nu

All of these approaches for their designated regions will produce relative errors less than 1e-16 using BigFloat arithmetic. However, all of them struggle a little bit when using Float64 in the whole routine.

1. The power series can be inaccurate when x>10. For large nu the log of the power series needs to be used
2. The asymptotic expansion also suffer from cancellation. They produce about 9e-15 relative errors for modest sized arguments and about 1e-13 for large arguments. 
3. I haven't tested the debye expansion but I think that will be easiest to get accurate.

For all other ranges for now I have filled using recurrence in the transition region x~v. For `besselj` downward recurrence is always stable. So when x<=nu I shift the order up and calculate using debye expansions and then use downward recurrence. When x>nu I will shift the order down and use the asymptotic expansion to compute then use forward recurrence which is stable when nu<=x. This method doesn't scale well for large orders because it relies on a simple multiplication factor for when the asymptotic are valid. On the other hand, the debye method is more uniform where simply shifting by a constant factor will allow for accurate results. This will allow for more comparable runtimes independent of order.

src/U_polynomials.jl
```diff
@@ -26,6 +26,23 @@ function Uk_poly_In(p, v, p2, ::Type{Float32})
...
+    u11 = -1 / 27636992044320430227456000 * evalpoly(p2, (15237265774872558064250728125, -2321657500166464779660536015625, 62011003282542082472252466220875, -676389476843440422173605288596087, 3926191452593448964331218647028194, -13704902022868787041097596217578170, 30589806122850866110941936529264950, -44801790321820682384740638703320750, 42936745153513007436411401865860625, -25963913760458280822131901737878125, 8997860461116953237934638500359375, -1363312191078326248171914924296875))
+    u12 = 1 / 39797268543821419527536640000 * evalpoly(p2, (120907703923613748239829527671875, -21882222767154197351962677311437500, 692277766674325563109617997687563750, -8958590476947726766450604043798559500, 62055079517573388459132793029571876461, -261201165596865827608687437905740780920, 714528665351965363868467348116538170900, -1314368459332124683504418467809129387000, 1642838631056253395899341314188134178125, -1378260730939829908037976894025628887500, 743739612850105971846901081692858843750, -233469939346545526651936774615688437500, 32426380464797989812768996474401171875))
```
heltonmc, line 42: These are also stored as `Int128` need to fix.

src/asymptotics.jl
```diff
@@ -0,0 +1,97 @@
...
+    μ = 4 * v^2
+    s0 = 1
+    s1 = (1 - μ) / 8
+    s2 = evalpoly(μ, (-0.1953125, 0.203125, -0.0078125))
+    s3 = evalpoly(μ, (1.0478515625, -1.1591796875, 0.1123046875, -0.0009765625))
+    s4 = evalpoly(μ, (-11.466461181640625, 13.1358642578125, -1.71624755859375, 0.0469970703125, -0.000152587890625))
+    s5 = evalpoly(μ, (211.27614974975586, -246.8455924987793, 37.067405700683594, -1.5151596069335938, 0.017223358154296875, -2.6702880859375e-5))
+    s6 = evalpoly(μ, (-5892.0457146167755, 6965.548173427582, -1128.6143367290497, 56.11658954620361, -1.0105445384979248, 0.005837917327880859, -5.0067901611328125e-6))
+
+    if x < 18*v
```
heltonmc, line 12: Need to figure out better way to select for how many terms here.
- 30 terms gives ~5e-16 relative error when x > 1.6*nu
- 20 terms gives ~2e-16 relative error when x > 2*nu
- 15 terms gives ~5e-18 relative error when x > 3*nu
- 10 terms gives ~3e-18 relative error when x > 5*nu

heltonmc:
I've fixed most of the tests here. An update on the algorithm

1. For x < 4.0 the power series can be used without numerical cancellation.
2. For x > 1.8*nu the asymptotic expansion can be used (given that x > 20.0)
3. For 2*x < nu the debye expansion can be used for moderate orders (50-1000) for very large orders it is accurate roughly with just a single shift (x < nu - 500)

For all other regions (4.0 < x < 20.0 and x ~ nu) it is best to shift the order upwards and use downward recurrence using starting values with the debye expansion. Downward is always stable for `besselj`.

Now given that this is kind of messy right now so need to go through and make sure all the branches are better defined.

src/besselj.jl
```diff
@@ -150,3 +150,203 @@ function besselj1(x::Float32)
...
+    T = eltype(x)
+    S = promote_type(T, Float64)
+    x = S(x)
+
+    vmx = fma(v,v, -x^2)
+    vdx = v/x
+    b = sqrt(vmx)
+
+    n = v * log(vdx + sqrt(vdx^2 - 1)) - b
+    coef = SQ1O2PI(S) * exp(-n) / sqrt(b)
```
heltonmc, line 272: `coef` has some floating point errors that needs to be calculated more accurately in F64 for better accuracy

heltonmc:
I think last issue to get sorted out is how to best select number of terms in the asymptotic expansion.
```julia

function _α_αp_asymptotic(v, x::Float64)
    xinv = inv(x)^2
    μ = 4 * v^2
    xnu = x / v
   
    N = 11#ceil(Int, 27.7479 / xnu + 22.3588 / xnu^2 + 4.5) # formula to determine number of terms needed
    d = -muladd(2, N, -3)

    s = evalpoly.(μ, _α_αp_POLYS(Float64)[1:N])

    αp = s[end]
    α = αp / d
    for i in N-1:-1:1
        d += 2
        αp = muladd(xinv, αp, s[i])
        α = muladd(xinv, α, s[i]/d)
    end

    return α*x, αp
end

const _α_αp_POLYS(::Type{Float64}) = (
    (1.0, 0.0),
    (0.125, -0.125),
    (-0.1953125, 0.203125, -0.0078125),
    (1.0478515625, -1.1591796875, 0.1123046875, -0.0009765625),
    (-11.466461181640625, 13.1358642578125, -1.71624755859375, 0.0469970703125, -0.000152587890625),
    (211.27614974975586, -246.8455924987793, 37.067405700683594, -1.5151596069335938, 0.017223358154296875, -2.6702880859375e-5),
    (-5892.0457146167755, 6965.548173427582, -1128.6143367290497, 56.11658954620361, -1.0105445384979248, 0.005837917327880859, -5.0067901611328125e-6),
    (231884.6359563172, -276225.9131319225, 46889.68168082833, -2607.1347218453884, 59.29277476668358, -0.5644365847110748, 0.0018794238567352295, -9.834766387939453e-7),
    (-1.225106763286615e7, 1.4670878180751745e7, -2.5692206274838205e6, 153362.43350451812, -4001.4928083932027, 49.41718461737037, -0.2788656409829855, 0.0005833245813846588, -1.9976869225502014e-7),
    (8.369564497799761e8, -1.0061560971482614e9, 1.80175839935865e8, -1.1293844119552301e7, 322272.3393258663, -4655.44006793818, 34.77845393284224, -0.12591543863527477, 0.0001761710154823959, -4.1618477553129196e-8),
    (-7.181611223268327e10, 8.659091218989731e10, -1.576853447634526e10, 1.0243675365394899e9, -3.1127612822052997e7, 499001.19046756154, -4427.281508191307, 21.557856538012857, -0.053067848184582544, 5.209072696743533e-5, -8.843926480039954e-9),
    (7.562132599676877e12, -9.139342212003133e12, 1.686263984004851e12, -1.1257747938365613e11, 3.584384568770644e9, -6.188986771037158e7, 616601.7257690447, -3609.808395761912, 12.103919447801218, -0.021179858537379914, 1.5144118606258417e-5, -1.909484126372263e-9),
    (-9.588236853265692e14, 1.161016591203577e15, -2.164822377227822e14, 1.4768015151140877e13, -4.8748767898914606e11, 8.900147704533855e9, -9.641151011038888e7, 640028.5410983711, -2606.385824929625, 6.277422845836725, -0.008087900592073538, 4.342405588886322e-6, -4.176996526439325e-10),
    (1.4409993748135704e17, -1.747635582560889e17, 3.286944341694018e16, -2.2818860422499875e15, 7.75248528894739e13, -1.4783938404512664e12, 1.7064572063331423e10, -1.2415577730993801e8, 578046.6774168271, -1705.2186748496474, 3.0504712266430722, -0.0029777336087200013, 1.2309086638850886e-6, -9.237588471933122e-11),
    (-2.5330554576498356e19, 3.0761765051109855e19, -5.827690025164462e18, 4.1047854971841363e17, -1.4277576770556304e16, 2.819724813029272e14, -3.4214400566025547e12, 2.670056925037507e10, -1.3717353508462998e8, 465890.6725849346, -1028.1881576107903, 1.4039031226739573, -0.0010631539871280848, 3.4554354377824836e-7, -2.061961712485072e-11),
    (5.149237018485509e21, -6.260437088008504e21, 1.1933270317945127e21, -8.508441473311652e19, 3.0180281284560543e18, -6.1344355915029736e16, 7.750474619361342e14, -6.393706192980096e12, 3.543452904857713e10, -1.3358182185805681e8, 341472.1144099866, -578.7806110577233, 0.6169921485050097, -0.00036976968916281605, 9.619824624385043e-8, -4.639413853091412e-12),
    (-1.1986350446061028e24, 1.458736721654471e24, -2.795293505867967e23, 2.0139038975089457e22, -7.262922096794037e20, 1.5122633519520676e19, -1.975555191650512e17, 1.7050149800082885e15, -1.0037673890747877e13, 4.1025284625807785e10, -1.1700278886842367e8, 230919.39384718135, -307.22947128726474, 0.260626935759175, -0.0001257334987819726, 2.658906071879852e-8, -1.051117201091023e-12),
    (3.167717078292175e26, -3.8584266638650446e26, 7.4277307068246446e25, -5.399729057323595e24, 1.9752248974202056e23, -4.197842582814349e21, 5.640147455618286e19, -5.0538188484912346e17, 3.1258357574848365e15, -1.3629748027461303e13, 4.233000035928751e10, -9.36296364559977e7, 145724.23973885347, -155.01792293897955, 0.10637041031529318, -4.191933776256961e-5, 7.303030923530292e-9, -2.3959289142515966e-13),
    (-9.432675683095812e28, 1.1498154949132059e29, -2.2224123014148547e28, 1.6283472309548362e27, -6.03034763221407e25, 1.3043849059659297e24, -1.7950710383376238e22, 1.6602291470434327e20, -1.0700309670904285e18, 4.92026735148727e15, -1.6365119263912514e13, 3.957355516860315e10, -6.930551343828379e7, 86599.59532013819, -74.82883257079231, 0.042123857348712865, -1.3735762526818516e-5, 1.994760566671595e-9, -5.490670428493242e-14),
    (3.1436617397720927e31, -3.8346121964186915e31, 7.438172213027661e30, -5.487658176284203e29, 2.0543110479331377e28, -4.512190593153004e26, 6.33949148154985e24, -6.024489803885022e22, 4.020727203383251e20, -1.9328978639294188e18, 6.802550072768408e15, -1.767751414955178e13, 3.3964900375451138e10, -4.792703592867111e7, 48822.71795456563, -34.73905348751307, 0.01624286720956346, -4.43222001918643e-6, 5.421693881235296e-10, -1.2642991118241018e-14),
    (-1.1655823449858069e34, 1.4226232333293694e34, -2.7683048917097874e33, 2.0549175959945544e32, -7.766193794571668e30, 1.7289192571208397e29, -2.473355129610242e27, 2.406335820680262e25, -1.6548511836241395e23, 8.261988206301513e20, -3.048956971903377e18, 8.409146358453666e15, -1.7416884938569994e13, 2.7048317602344254e10, -3.1216507081066556e7, 26272.21159305084, -15.578788452840763, 0.0061163849392160095, -1.4107081622121022e-6, 1.4671084930995856e-10, -2.9236916960932355e-15),
    (4.7820456670665866e36, -5.839767806978455e36, 1.139609569657776e36, -8.505736729879162e34, 3.241923388548771e33, -7.303696541971675e31, 1.061598557628623e30, -1.0542835823932327e28, 7.4416439961740466e25, -3.8383112218012e23, 1.4749530840065974e21, -4.277160387520148e18, 9.428034079487212e15, -1.582775791680076e13, 2.0161618573658318e10, -1.9279964972155876e7, 13562.78784549061, -6.773533616718505, 0.0022547380647248, -4.4351435924730097e-7, 3.9542564728199e-11, -6.787141437359297e-16),
    (-2.1603839510459968e39, 2.639521237493071e39, -5.164169654060501e38, 3.873385769490781e37, -1.4875551809791352e36, 3.3870740530372218e34, -4.993086633044377e32, 5.049452384590897e30, -3.646484825050138e28, 1.934929035289944e26, -7.699788293506606e23, 2.3306500988206443e21, -5.414879939834449e18, 9.69892178268169e15, -1.3389590376401615e13, 1.4169012254830626e10, -1.1355350548005087e7, 6745.941285011119, -2.8643019266046954, 0.0008154154591207295, -1.3789599324701242e-7, 1.061958568923219e-11, -1.5810954484757452e-16),
    (1.0699708275991454e42, -1.3078533633276038e42, 2.5647480608865752e41, -1.9322501955238208e40, 7.471528599453178e38, -1.7175047656287374e37, 2.56398527252613e35, -2.6350896428114214e33, 1.9417708082703794e31, -1.0563755426904998e29, 4.333903992576568e26, -1.3614388053852227e24, 3.3089693719445074e21, -6.261128573879847e18, 9.243346554045246e15, -1.0625592945166727e13, 9.4457596395905e9, -6.4083469576330995e6, 3244.624388704194, -1.1811472686414501, 0.0002898188382446864, -4.2443950630387665e-8, 2.84271079789386e-12, -3.694951319807448e-17),
    (-5.786059033524315e44, 7.075318200534661e44, -1.3904375539321083e44, 1.051770598467428e43, -4.0921342782576984e41, 9.487906954286141e39, -1.4325415965940442e38, 1.493677625028016e36, -1.1206631946323043e34, 6.2329465039514265e31, -2.626782977025989e29, 8.52403241555426e26, -2.1544180262833415e24, 4.273241060880315e21, -6.678193343397941e18, 8.226793217207085e15, -7.9615210223039795e12, 6.004451011853662e9, -3.4795218728555855e6, 1513.831326319914, -0.4760701870709445, 0.00010139500470517733, -1.2944568154864108e-8, 7.586889731852131e-13, -8.660042155798706e-18),
    (3.4037481878842518e47, -4.163722422572948e47, 8.198385202264364e46, -6.22435990386475e45, 2.4353618503436353e44, -5.690731110001809e42, 8.680616408346572e40, -9.169497515863974e38, 6.991529767882026e36, -3.9660440566717967e34, 1.7117897755074633e32, -5.716342746132164e29, 1.495186253411991e27, -3.089698915505606e24, 5.071094031947313e21, -6.62564031827467e18, 6.884682491993269e15, -5.663687113496129e12, 3.6558083666029415e9, -1.8241160174217927e6, 687.0067933018984, -0.1879264018835715, 3.496541491784091e-5, -3.914759734732158e-9, 2.0193378048364477e-13, -2.035109906612696e-18),
    (-2.170796074547977e50, 2.6563886780963143e50, -5.239744551853566e49, 3.9915243203539826e48, -1.5697710844957169e47, 3.694226454576932e45, -5.687790414754595e43, 6.07923257088654e41, -4.703231744077508e39, 2.715669597352839e37, -1.1973880447446774e35, 4.101798034247621e32, -1.1059082744278117e30, 2.3689970250205793e27, -4.057770223206878e24, 5.577674944614657e21, -6.157880219725851e18, 5.448993990587677e15, -3.8435237604702446e12, 2.1401262084993436e9, -926134.0995426074, 303.9782133922326, -0.07278018905169194, 1.1898982367956518e-5, -1.1748038677566303e-9, 5.361217221358988e-14, -4.794249299231832e-19),
    (1.4962552090981924e53, -1.8315342344152888e53, 3.6186332587139524e52, -2.7651411770117913e51, 1.092593756883467e50, -2.5880088398643682e48, 4.0185482602383215e46, -4.341325948517365e44, 3.40330197791503e42, -1.996788871925338e40, 8.974814912978575e37, -3.1454308144632116e35, 8.712781142577153e32, -1.926817207371682e30, 3.4266421642703216e27, -4.923376108039163e24, 5.727723667496557e21, -5.393876507549732e18, 4.099268265682976e15, -2.498492707776039e12, 1.2086394541382809e9, -456609.83322422515, 131.40817769348385, -0.027695807224792045, 4.000233368100494e-6, -3.500503095889607e-10, 1.4200746208673407e-14, -1.1319755289852937e-19),
    (-1.1113660587118494e56, 1.360796163396552e56, -2.692648562464633e55, 2.0634410442908336e54, -8.188692555802185e52, 1.9512412867297912e51, -3.0534326199740573e49, 3.3310997303957132e47, -2.6429272553053182e45, 1.573358131379098e43, -7.195503085038196e40, 2.5742410447166102e38, -7.305470671594484e35, 1.6621868613473995e33, -3.056128428896728e30, 4.5656808004766417e27, -5.560280477035352e24, 5.525982611412455e21, -4.4762143769251487e18, 2.944057191090032e15, -1.561376913130651e12, 6.60436019199837e8, -219122.78889988337, 55.602561558610454, -0.010370022987617815, 1.3297453832052619e-6, -1.0361746982713131e-10, 3.753418608703549e-15, -2.6783349569741323e-20),
    (8.871640240590993e58, -1.0865691207568645e59, 2.153044379065757e58, -1.654292598675498e57, 6.591312178536168e55, -1.5792626981209235e54, 2.4890507915218916e52, -2.7398535977577976e50, 2.1978557435213864e48, -1.3258581553903244e46, 6.160064953128667e43, -2.24525395106052e41, 6.512651083776791e38, -1.520126754804397e36, 2.879357904563898e33, -4.453253892004961e30, 5.646779978807957e27, -5.882834513685185e24, 5.036061410449725e21, -3.535410901341911e18, 2.0262400502486065e15, -9.410029770358712e11, 3.500781492364049e8, -102565.9203518178, 23.065519747567176, -0.003825019881917803, 4.374396753566091e-7, -3.048458029283708e-11, 9.900967356259312e-16, -6.3495009755852275e-21),
    (-7.592029625744128e61, 9.300807898884993e61, -1.8453653700408117e61, 1.4213662513968141e60, -5.684267643368571e58, 1.3688721521734006e57, -2.1717210753568686e55, 2.410362055956752e53, -1.953160895382593e51, 1.1926329508677187e49, -5.621526725185766e46, 2.084010993835956e44, -6.16598654084696e41, 1.472798231204338e39, -2.8653790379578294e36, 4.5711759648399396e33, -6.008229949414824e30, 6.525582948238488e27, -5.863525845645105e24, 4.3559252317865593e21, -2.6682229171870643e18, 1.3409000748848645e15, -5.4845348768282446e11, 1.8042426705573353e8, -46913.0207873715, 9.39407807634952, -0.0013913710768178739, 1.4251320396079487e-7, -8.917800204478063e-12, 2.6068908049173686e-16, -1.5080064817014915e-21)
)
```
This function does what I want but it's type unstable. I think I have to set up `_α_αp_POLYS` differently as they are storing different size `NTuples` and then the type of `s` depends on `N`. Ideally, N could be determined by the formula I've given but can't get it to work. If N<=10 this function is type stable but for N>10 it is not.

@oscardssmith if you have any ideas on best to set this function up that would be helpful

heltonmc:
Ok I have now hopefully worked through all the edge cases where we must use recursion. There are a few places where accuracy could be improved that I've noted above. Mainly the asymptotic expansion could be computed more accurately and the power series has some larger errors when the order is large.

Overall, this approach can be significantly faster (especially for large orders). A particularly egregious example is...

```julia
julia> @benchmark SpecialFunctions.besselj(1.0e7, x) setup=(x=rand()+9990000.0)
BenchmarkTools.Trial: 95 samples with 1 evaluation.
 Range (min … max):  52.244 ms …  53.992 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     53.110 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   53.103 ms ± 346.777 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                     ▂▅▂  ▂ ██▂██▅ ▂▅▅▂▂  █                     
  ▅▁▁▅▅▅▁█▁▁▁▁▅██▅▁█▁███▁▁████████▅█████▁███▁▁▅▅▅▅██▅▁▁▁▁▅▁▁▁▅ ▁
  52.2 ms         Histogram: frequency by time           54 ms <

julia> @benchmark Bessels._besselj(1.0e7, x) setup=(x=rand()+9990000.0)
BenchmarkTools.Trial: 10000 samples with 737 evaluations.
 Range (min … max):  171.924 ns … 223.822 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     172.863 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   173.829 ns ±   3.753 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▄▆██▆▄▂    ▃▃▂▁ ▁▂▂▁                                         ▂
  ████████▆▅▆██████████████████▇▆▅▇▇▆▄▅▅▄▆▅▅▇▇▇▇▅▄▂▄▇▇▆▄▆▆▄▅▃▄▄ █
  172 ns        Histogram: log(frequency) by time        185 ns <

julia> Bessels._besselj(1.0e7, 9990000.0)
1.7206853104017583e-133

julia> SpecialFunctions.besselj(1.0e7, 9990000.0)
1.7206853100053732e-133


```
That is roughly 300,000x faster....... For small orders the difference is less but still significant
```julia
julia> @btime SpecialFunctions.besselj(1.0e2, x) setup=(x=9990.0)
  310.601 ns (0 allocations: 0 bytes)
0.006807635943512029

julia> @btime Bessels._besselj(1.0e2, x) setup=(x=9990.0)
  54.115 ns (0 allocations: 0 bytes)
0.006807635943512404
```

heltonmc:
I'm going to merge this as it provides accurate results. I have several other optimizations I'd like to consider but feel this thread is getting fairly long. I will leave `_besselj` unexported for now as I have a few other changes to the algorithm. Essentially the debye and asymptotic expansion can give both `besselj` and `bessely` in the same time which will be important for the hankel functions.

PR #23, opened 2022-07-19
optimize U-polynomials

heltonmc:
This seeks to eliminate some code overlap and optimize the evaluation of U-polynomials.

In general we need to compute both 

$$ \sum_{k=0}^{\infty}  \frac{U_k(p)}{\nu^k} \mathrm{and}  \sum_{k=0}^{\infty} (-1)^k  \frac{U_k(p)}{\nu^k}$$

to be able to calculate all of `besselj`, `bessely`, `besseli` and `besselk`.

The old function looked like...
```julia
function Uk_poly_Jn20(p, v, p2, ::Type{T}) where T <: Float64
    u0 = one(T)
    u1 = -evalpoly(p2, (0.125, -0.20833333333333334))
    u2 = evalpoly(p2, (0.0703125, -0.4010416666666667, 0.3342013888888889))
    u3 = -evalpoly(p2, (0.0732421875, -0.8912109375, 1.8464626736111112, -1.0258125964506173))
    u4 = evalpoly(p2, (0.112152099609375, -2.3640869140625, 8.78912353515625, -11.207002616222994, 4.669584423426247))
    u5 = -evalpoly(p2, (0.22710800170898438, -7.368794359479632, 42.53499874538846, -91.81824154324002, 84.63621767460073, -28.212072558200244))
    u6 = evalpoly(p2, (0.5725014209747314, -26.491430486951554, 218.1905117442116, -699.5796273761325, 1059.9904525279999, -765.2524681411817, 212.57013003921713))
    u7 = -evalpoly(p2, (1.7277275025844574, -108.09091978839466, 1200.9029132163525, -5305.646978613403, 11655.393336864534, -13586.550006434136, 8061.722181737309, -1919.457662318407))
    u8 = evalpoly(p2, (6.074042001273483, -493.91530477308805, 7109.514302489364, -41192.65496889755, 122200.46498301747, -203400.17728041555, 192547.00123253153, -96980.59838863752, 20204.29133096615))
    u9 = -evalpoly(p2, (24.380529699556064, -2499.8304818112097, 45218.76898136273, -331645.17248456355, 1.2683652733216248e6, -2.8135632265865337e6, 3.763271297656404e6, -2.998015918538107e6, 1.3117636146629772e6, -242919.18790055133))
[...]

PR #24, opened 2022-07-20
Output both bessely and besselj for large arguments

heltonmc:
This modifies the asymptotic expansion to export both `bessely` and `besselj` for large arguments. This allows for computation of both `bessely` and `besselj` in about the same time.

Computing just `besselj` on master.

```julia
julia> @benchmark Bessels.besselj_large_argument(10.0, x) setup=(x=100.0+rand())

BenchmarkTools.Trial: 10000 samples with 986 evaluations.
 Range (min … max):  51.470 ns … 87.352 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     51.637 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   52.009 ns ±  2.329 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ██▅       ▃▂▁                                               ▂
  ███▁▄▄▁▁▃▆███▆▆▅▄▄▅▆▇▆▅▇▅▄▃▅▅▅▄▄▃▄▄▃▅▃▃▅▄▁▃▄▃▃▁▃▃▁▁▃▄▄▃▅▄▁▄ █
  51.5 ns      Histogram: log(frequency) by time        61 ns <

```

and computing both `besselj` and `bessely` on this PR
```julia
julia> @benchmark Bessels.besseljy_large_argument(10.0, x) setup=(x=100.0+rand())
BenchmarkTools.Trial: 10000 samples with 985 evaluations.
 Range (min … max):  54.583 ns … 85.532 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     54.871 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   55.354 ns ±  2.262 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅█▆▃     ▂▂                                                 ▁
  ████▄▄▁▄▇███▇▆▆▆▇▇▇▆▆▇▆▆▆▅▆▆▄▅▅▃▄▃▃▆▅▃▅▄▅▆▅▇█▁▅▅▃▅▃▅▄▄▅▄▅▃▄ █
  54.6 ns      Histogram: log(frequency) by time      66.2 ns <

```

(I'm going to try to keep these PRs smaller)

### DIFF
--- a/src/asymptotics.jl
+++ b/src/asymptotics.jl
@@ -1,3 +1,26 @@
+function besseljy_large_argument(v, x::T) where T
+    # gives both (besselj, bessely) for x > 1.6*v
+    α, αp = _α_αp_asymptotic(v, x)
+    b = SQ2OPI(T) / sqrt(αp * x)
+
+    # we need to calculate sin(x - v*pi/2 - pi/4) and cos(x - v*pi/2 - pi/4)
+    # For improved accuracy this is expanded using the formula for sin(x+y+z)
+
+    S, C = sincos(PIO2(T) * v)
+    Sα, Cα = sincos(α)
+
+    CMS = C - S
+    CPS = C + S
+
+    s1 = CMS * Cα
+    s2 = CPS * Sα
+
+    s3 = CMS * Sα
+    s4 = CPS * Cα
+
+    return SQ2O2(T) * (s1 + s2) * b, SQ2O2(T) * (s3 - s4) * b
+end
+
 function _α_αp_asymptotic(v, x::Float64)
     if x > 5*v
         return _α_αp_poly_10(v, x)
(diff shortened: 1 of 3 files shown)


heltonmc:
This is fairly straightforward so will go ahead and merge. I can then bridge with #23 to give generic `bessely` implementations. We may want to come back to https://github.com/heltonmc/Bessels.jl/blob/3b3c2e9985ad06b7da4f8c9efbc20b3b2ddcfa41/src/asymptotics.jl#L9-L19 to see if this can be done more accurately but we are also limited by the asymptotic nature of evaluating for very large `x` values which limits some of the accuracy

PR #28, opened 2022-08-02
Clean up besselj(nu,x) and add docs

heltonmc:
I've added more docs to `besselj` and other internal methods.

I've also cleaned up the `besselj` implementation to be more consistent with `bessely` while removing several branches due to using the Hankel expansion instead of using either forward or backward recurrence. I think this helps with branch prediction and speeding up the overall routine. Accuracy is unchanged. I've also removed the `loggamma` implementation from SpecialFunctions.jl so this function is entirely in Julia now.

```julia
julia> @benchmark Bessels.besselj(y, x) setup=(x=rand()*300+20.0; y=rand()*300)
BenchmarkTools.Trial: 10000 samples with 963 evaluations.
 Range (min … max):   84.805 ns … 580.607 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     146.028 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   178.862 ns ± 115.707 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █ ▆▅     ▆▇   ▅▄                ▂▂▁▂▂▂▂▁▂▂▂▂▂▂▂▁▁▂▂▁▂▁▁▁▁▁▁▁▁ ▂
  █▆██▇▅▄▄▃██▇▆▄██▆▄▃▄▃▁▁▁▁▁▁▁▁▁▁▁█████████████████████████████ █
  84.8 ns       Histogram: log(frequency) by time        475 ns <

julia> @benchmark SpecialFunctions.besselj(y, x) setup=(x=rand()*300+20.0; y=rand()*300)
BenchmarkTools.Trial: 4171 samples with 535 evaluations.
 Range (min … max):  207.865 ns … 7.374 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):       1.464 μs             ┊ GC (median):    0.00%
 Time  (mean ± σ):     2.237 μs ± 1.453 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄▅ ▂   ▁▄█▆▅▁▁    ▄▆▅▁▂▂▂▁  ▃▃▂     ▃▄▄▃▁▁                  ▂
  ████▃▃▅████████▃▃▅█████████▇███▇▃▄▁▇███████▅▆▆▅▃▁▁▁▃▄▃▁▇███ █
  208 ns       Histogram: log(frequency) by time      6.83 μs <

```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,11 +3,8 @@ uuid = "0e736298-9ec6-45e8-9647-e4fc86a2fe38"
 authors = ["Michael Helton <heltonmc@protonmail.com> and contributors"]
 version = "0.1.0"
 
-[deps]
-SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
-
 [compat]
-julia = "1.5"
+julia = "1.6"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/src/Bessels.jl
+++ b/src/Bessels.jl
@@ -1,7 +1,5 @@
 module Bessels
 
-using SpecialFunctions: loggamma
-
 export besselj0
 export besselj1
 export besselj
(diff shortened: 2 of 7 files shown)


heltonmc:
I'm going to merge this before #29 and deal with any conflicts.

Issue #17, opened 2022-05-09
Automatic differentiation (AD) not supported


heltonmc:
Currently, implementations don't support automatic differentiation with packages such as `ForwardDiff.jl`. 

Probably best to give ChainRules such as in https://github.com/JuliaMath/SpecialFunctions.jl/blob/master/src/chainrules.jl or take those implementations directly as this package matures.

oscardssmith:
Also, you should look at https://github.com/cgeoga/BesselK.jl. It allows for autodiff of `besselk` with respect to the order which is pretty cool. Also, it apparently uses some fancyness to automatically generate some of the polynomial coefficients rather than requiring hand-derivation.

heltonmc:
Just making a note of this here. There was some interest from from Slack about getting derivatives to work with respect to both argument and order here. SpecialFunctions.jl only provides rules for argument derivatives. BesselK.jl shows that this is possible for orders with some care.

This should be possible once we get the arbitrary orders done, but this seems like functionality people are interested in so I'll work on this after we finish the arbitrary orders. Though, historically `ForwardDiff.jl` gives me fits...

cgeoga:
Just was passing through and was curious how stuff here was going, and this is pretty neat! From a little tinkering it seems like when the argument `x` is small these might be a bit slow to converge. I would have thought that so many `exp` and `cosh` calls would just totally kill performance, but even as provided its within, like, a factor of 4 of what I did a lot of grinding on. I also would have thought that the `tmp` variable would be an issue because it is a "big*tiny" operation, but the accuracy is also really impressive. Just with standard identities for cosh(x+y) and sinh(x+y) I wrote this version that does one of the optimizations I assume you had in mind:
```
function _besselk(v, x::T) where T
    MaxIter = 5000
    tol = eps(T)
    h = T(0.1)
    t = h
    out = 0.5*exp(-x)
    # pre-compute a few nice quantities:
    (cosh_t,  sinh_t)  = (cosh(t),   sinh(t))
    (cosh_vt, sinh_vt) = (cosh(v*t), sinh(v*t))
    (cosh_h,  sinh_h)  = (cosh(h),   sinh(h))
    (cosh_vh, sinh_vh) = (cosh(v*h), sinh(v*h))
    for _ in 1:MaxIter
        tmp  = cosh_vt*exp(-x*cosh_t)
        out += tmp
        tmp / out < tol && break
        # update the quantities, t -> t+h, vt -> v(t+h):
[...]

Issue #25, opened 2022-07-26
Half-integer order for `besselk`


cgeoga:
What do you think about something like this to handle half-integer orders of `besselk`? It uses the fact that the asymptotic expansion terminates and is exact for half-integer orders.

```
function _besselk_vhalfint(v, x) 
  v<zero(v)  && return _besselk_vhalfint(-v, x)  
  (v == 1/2) && return sqrt(pi/(2*x))*exp(-x)
  (v == 3/2) && return sqrt(pi/(2*x))*exp(-x)*(one(x) + inv(x))
  b0 = sqrt(pi/(2*x))*exp(-x)                    # v=1/2
  b1 = sqrt(pi/(2*x))*exp(-x)*(one(x) + inv(x))  # v=3/2
  _v = 3/2
  twodx = 2/x
  b2 = twodx*_v*b1 + b0
  for _ in 1:Int(floor(v-2)) # standard up-recurrence
    _v += 1
    b0  = b1
    b1  = b2
    b2  = muladd(twodx*_v, b1, b0)
  end
  b2
end
```

You can't pass AD through this w.r.t. `v`, of course, and this was in my experience the hardest derivative to get and is why I ended up coding up the very expensive Temme routine, which is by far the slowest routine for us. But if you wanted something in the mean time, this has to be pretty competitively fast except for outrageously large `v`.

heltonmc:
I'm not for sure if there is another function I'm missing but that allocates because it is just misspelled (missing an `_` in front) so it can't infer the correct type of that return. Using `@code_warntype` is always helpful for those kind of weird things.

One thing I would be concerned about is just the amount of branches. For the `v<zero(v)` I would probably just say v = abs(v) since `K_{-v}(x) = K_{v}(x)`?

Though there should be a way to eliminate those branches for constants like that... see comment https://github.com/JuliaMath/SpecialFunctions.jl/issues/178#issuecomment-1128113294 and related notes

cgeoga:
Okay, well that's embarrassing about the `_`. I actually did look at the output of `@code_warntype`, but I did not understand it well enough to see that that was the problem. Oof. I've edited the top post so that code works correctly on a copy+paste. Also, makes sense about `abs(v)`, although `-v` would give the same output in this case.

With regard to the branches, I'm not sure how they could really be removed and if there's something in that issue explaining how I don't think I understand it. Even with manual methods, how would you eventually avoid somehow checking if `v` is one of the two base cases?

[...]

PR #29, opened 2022-08-03
Besselj and Besselk for all integer and noninteger postive orders

heltonmc:
Implementation of `besselk` and `besseli` for all real orders. Still have several things to finish and touch up but these methods so far seem to work

`besseli` is very straightforward. Use the uniform asymptotic expansion when either `x` or `nu` is large (x>35, nu>25) and then the power series elsewhere... now I also have the asymptotic expansion when x is large in there that is significantly faster but may just remove so the whole function has just two branches... decision point for later..

`besselk` is obviously not straightforward. However, the same uniform asymptotic expansion applies when either `x` or `nu` is large (x>35, nu>25) so use that. Now we could also use an asymptotic expansion with respect to `x` here as well (see https://github.com/heltonmc/Bessels.jl/issues/17#issuecomment-1197096678), but again at the cost of an extra branch.

So the rest comes down to what do we do when both x < 35 and nu < 25... The power series form for `besselk` is accurate when x < 2 or nu > 1.6x - 1.0 so we could use that but that doesn't help much for x > 2 and adds an extra branch. We could use numerical integration here from small orders and then use forward recurrence (see https://github.com/heltonmc/Bessels.jl/issues/17#issuecomment-1177801831) and subsequent optimizations.... We could also use the hypergeometric series (see https://github.com/heltonmc/Bessels.jl/issues/17#issuecomment-1196836506) for small orders then use a continued fraction approach to generate k_[nu+1} and then use forward recurrence to fill everything else..... we could also use some type of polynomial approximation (Chebyshev) but I found that this only worked well for a limited range.

I think the simplest method is to take advantage of the fact that the `besseli` power series is very accurate for nu<25 for any x value.... the advantage is we can also generate I_{nu+1} very easily so we can get very accurate estimates of  I_{nu+1}/I_{nu} for any value of nu in this range. Now, we can also use a continued fraction approach to get K_{nu+1}/K_{nu}  (see https://github.com/heltonmc/Bessels.jl/issues/17#issuecomment-1195726642). So using the two ratios we can use the Wronskian to get K_{nu} very simply with a lot of code reuse.

Here are some benchmarks for this method.
```julia
julia> @benchmark Bessels.besselk_continued_fraction(15.2, 17.0)
BenchmarkTools.Trial: 10000 samples with 700 evaluations.
 Range (min … max):  179.583 ns … 276.309 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     179.881 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   180.464 ns ±   3.142 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▆█▇▄▁                            ▂▃▁                         ▁
  ███████▇▆▆▆▅▅▅▅▄▃▄▄▄▄▅▅▅▄▄▄▄▄▃▃▂▄▇███▆▃▃▅▂▄▄▄▂▄▂▃▄▃▅▅▅▆▆▇▆▆▅▅ █
  180 ns        Histogram: log(frequency) by time        187 ns <

julia> @benchmark SpecialFunctions.besselk(15.2, 17.0)
BenchmarkTools.Trial: 10000 samples with 198 evaluations.
 Range (min … max):  435.606 ns … 782.409 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     441.081 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   450.682 ns ±  26.869 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▃██▆▃▃▁▃▄▂▁▁       ▁                                          ▂
  █████████████████▇▇████▇▇▆▇▇▇▇█▇▇▆▇█▇▆▇▇▇▇▆▆▆▇▆▇▆▆▅▅▄▅▅▅▅▅▅▅▄ █
  436 ns        Histogram: log(frequency) by time        571 ns <

```
So almost 2.5x faster than SpecialFunctions....

O and and here is a plot of the relative errors compared to ArbNumerics.jl
Screen Shot 2022-08-03 at 12 28 49 PM

With a maximum error of..
```julia
julia> maximum(out)
1.33399240780568911954050188375962391214648355184706163492721e-15
```
For reference this more accurate than SpecialFunctions.jl which had a maximum error of 5e-15.


heltonmc:
Some general things going forward.... Do we want to have another branch for large x that would be faster at the cost of more branches? Unsure honestly. We can look more at @cgeoga large argument expansion https://github.com/heltonmc/Bessels.jl/issues/17#issuecomment-1197096678. I have one for `besseli` that is really fast and accurate in the appropriate range

oscardssmith:
That looks great!

cgeoga:
Hey, sorry to have disappeared for a few days! I'm about to enter a pretty swamped period and will probably be slower for the next week or so.

Those times for the intermediate range are very impressive to me. I played a bit more with the Miller recurrence, but I think to get to the tolerances you are going for here it would just be too expensive. With regard to very large `x`, I'm surprised about the standard asymptotic expansion not being fast enough. With my random choice of `(v,x) = (0.3,38.1)`, if I crank up the order to `10` instead of `6` that was in your test, even without the exponential improvement I get `1 - (...)/(...) = 8.8e-16` or so at a speed cost `16 ns`. That suggests a higher `x` cutoff than I expected, but in my experience the uniform asymptotic expansion is much slower to converge as `x` gets larger when `v` is small. With that said, though, the AMOS library (figure 2 [here](https://dl.acm.org/doi/pdf/10.1145/7921.214331)) does not seem to have a special branch for this. The Temme 1975 paper does use one, though.

I like the idea of the code re-use with the `besseli` series and the Wronskian. That is a strategy that I missed but seems very simple and elegant. I wish I had taken the continued fraction thing more seriously earlier in the project and given that version a whack.

 With regard to extra branches, I would think there are enough ns on the table where it would make sense to have several for various specialized cases, like when `v = k + 1/2` for integer `k` (although I'm biased about this obviously). I would also think that a power series approach might be _way_ faster for, say, `x<0.1`. I don't really know what the computational physics applications are and what kinds of `(v,x)` you typically use, but in my applications I might hit that case a lot.

heltonmc:
Yes - the miller recurrence is definitely something else to consider! I think that is what Amos is using? So I would imagine that we might be able to slightly faster than what they have but similar accuracies. I do like this approach as the amount of code is much smaller and easier to optimize and diagnose branch cutoffs.

But to your point. The continued fraction approach is much slower when x < 1 so we should use a power series here for sure. I'll do that now.

For the large argument, I'm very happy to include that. Maybe once I merge this you could submit a PR with the large argument branch that you wrote? We can then integrate I think. I do like to lean on the asymptotic expansions just because they are so optimized (and I use them for every single routine) but definitely when x>>nu a different expansion will be better so let's do that.

I think for specific nu like 1/2 it would be worth having an optimized branch. Maybe as a separate PR after sort this all out. I'm also going to work on nu =1/3 because we can use those for Airy functions which are needed as well.... but in time.

heltonmc:
Here is a plot of the relative errors using the uniform asymptotic expansion (with 10 U polynomials). So it looks like it converges fairly well even when nu is small. 

Screen Shot 2022-08-03 at 1 48 08 PM

cgeoga:
Fantastic. This code is sort of on a pedestal for me and I'm nervous to submit an un-clean PR, but I'm very happy to give it a shot. Do you have a series code already? If not, our series code is pretty decent, I think, and I could try to clean it up and submit it for consideration as well in a separate issue or something.

Once this merges I'll definitely submit an asymptotic expansion PR. Do you want the exponential correction in that? The way I wrote it in our code it uses the maximum possible order, but maybe it would be faster for me to hard-code just using the leading term in it or something, if you wanted. Happy to submit it that way and potentially prune it off if the speed cost is too high. 

I could also add the half-integer order code from #25 in that same PR if you'd like.

Also, funny about that plot---in my mind, an order of, like, `3` is "large" because of my application. I keep forgetting that you want to work with an order of `50` and stuff.

heltonmc:
So I'd probably pose the following branches.

1. nu > 25.0 || x > 35.0
    - use the uniform asymptotic expansion
2. x < ~2.0
    - use the power series
3.  else
   - use continued fraction with wronskian
   - inu and inum1 are generate simultaneously with power series

With the option of a 4th branch for asymptotic x >> nu.

That sounds pretty reasonable to me...?

heltonmc:
Definitely don't feel that way. The code is very good so I think we can just drop in and clean some stuff up! But let's plan on that when you have time submit the asymptotic expansion for large x and we can go through it. The exponential correction sounds reasonable but we should have a pretty specific cutoff where we want to use it and set the order to a specific value for the entire branch. 

I do not have a power series for `besselk` that I am super happy with so it might be great to just reuse what you have. I would like to probably include that in this PR though. I can coauthor a commit if you want with that implementation here?

But yes if you want half orders please go ahead and submit that (preferably as a separate PR)!

Haha about the orders! In my research I deal with a lot of infinite series over the order so we need calculations for very high (nu>1e4) numbers....

cgeoga:
I appreciate that! I see you committed the code from the repo directly, but here is a slightly cleaned up version that I think is a slightly better style fit:
```
function _besselk_ser(v, x)
  (maxit, tol) = (100, 1e-15)
  T = promote_type(typeof(x), typeof(v))
  # precompute a handful of things:
  xd2  = x/T(2)
  xd22 = xd2*xd2
  half = one(T)/T(2)
  # (x/2)^(±v). Writing the literal power doesn't seem to change anything here,
  # and I think this is faster:
  lxd2     = log(xd2)
  xd2_v    = exp(v*lxd2)
  xd2_nv   = exp(-v*lxd2)
  # use the gamma function a couple times to start:
  gam_v    = gamma(v)
  gam_nv   = gamma(-v)
  gam_1mv  = -gam_nv*v # == gamma(one(T)-v)
  gam_1mnv = gam_v*v   # == gamma(one(T)+v)
  (gpv, gmv) = (gam_1mnv, gam_1mv)
  # One final re-compression of a few things:
  _t1 = gam_v*xd2_nv*gam_1mv
  _t2 = gam_nv*xd2_v*gam_1mnv
  # A couple series-specific accumulators:
  (xd2_pow, fact_k, floatk, out) = (one(T), one(T), zero(T), zero(T))
  for _j in 0:maxit
    t1   = half*xd2_pow
    tmp  = _t1/(gmv*fact_k)
    tmp += _t2/(gpv*fact_k)
    term = t1*tmp
    out += term
    abs(term/out) < tol && return out
    # Use the trick that gamma(1+k+1+v) == gamma(1+k+v)*(1+k+v) to skip gamma calls:
    (gpv, gmv) = (gpv*(one(T)+v+floatk), gmv*(one(T)-v+floatk)) 
    xd2_pow *= xd22
    fact_k  *= (floatk+1)
    floatk  += one(T)
  end
  throw(error("$maxit iterations reached without achieving atol $tol."))
end
```

Note that the modified version in our repo is not `exp(x)*besselk(v,x)`, it is `(x^v)*besselk(v, x)` (which shows up in the Matern covariance), so that probably isn't useful to you here and I cut it out.

The issue I'm having, though, is that my `1 - (...)/(...)` errors are really about `2e-14` and I can't do much better. Does anything here stand out to you as problematic?

src/besselk.jl
```diff
@@ -271,3 +271,52 @@ function besselk_ratio_knu_knup1(v, x::T) where T
...
+    xd2_nv   = exp(-v*lxd2)
+  end
+  gam_v    = gamma(v)
+  gam_nv   = gamma(-v)
+  gam_1mv  = -gam_nv*v # == gamma(one(T)-v)
+  gam_1mnv = gam_v*v   # == gamma(one(T)+v)
+  xd2_pow  = oneT
+  fact_k   = oneT
+  floatk   = convert(T, 0)
+  (gpv, gmv) = (gam_1mnv, gam_1mv)
```
oscardssmith, line 301: this doesn't do anything?

cgeoga: Maybe I'm misunderstanding the question, but those values are used inside the loop. I moved them closer to the loop declaration in the version posted here.

src/besselk.jl
```diff
@@ -271,3 +271,52 @@ function besselk_ratio_knu_knup1(v, x::T) where T
...
+    e2v      = exp2(v)
+    xd2_v    = (x^(2*v))/e2v
+    xd2_nv   = e2v
+  else
+    lxd2     = log(xd2)
+    xd2_v    = exp(v*lxd2)
+    xd2_nv   = exp(-v*lxd2)
+  end
+  gam_v    = gamma(v)
+  gam_nv   = gamma(-v)
```
oscardssmith, line 295: this can use the reflection formula.

src/besselk.jl
oscardssmith: `gpv` and `gmv` are used after this, but `gam_1mnv`  and `gam_1mv` aren't, so this is just a renaming.

src/besselk.jl
heltonmc: fixed this (may want to also include a check that v can not be an integer but we check in the top level)

src/besselk.jl
heltonmc: Removed this extra assignment

src/besselk.jl
```diff
@@ -272,51 +274,44 @@ function besselk_ratio_knu_knup1(v, x::T) where T
...
+    half = one(T) / 2
+    # (x/2)^(±v). Writing the literal power doesn't seem to change anything here,
+    # and I think this is faster:
+    lxd2 = log(xd2)
+    xd2_v = exp(v*lxd2)
+    xd2_nv = exp(-v*lxd2)
+    # use the gamma function a couple times to start:
+    gam_v = gamma(v)
+    xp1 = abs(v) + one(T)
+    gam_nv = π / sinpi(xp1) / _gamma(xp1)
```
oscardssmith, line 294: `π / (sinpi(xp1) * _gamma(xp1))`

src/besselk.jl
oscardssmith: If this is an internal function, I would just document it. `isinteger` isn't totally free.

src/besselk.jl
heltonmc: Fixed.

src/besselk.jl
heltonmc: Yes it's internal. We dispatch to a different method for integer orders at top level. Will add notes to reflect that

src/besselk.jl
heltonmc: I was dumb and didn't remove the call to _gamma. This is fixed now `gam_v*v =      gam_nv = π / sinpi(xp1) / _gamma(xp1) `

heltonmc: Is a division usually faster than an exponent?
```julia
logz = log(z)
    xd2_v = exp(v*logz)
    xd2_nv = inv(xd2_v)#exp(-v*logz)
```

oscardssmith: yes, by roughly a factor of 2.

src/besselk.jl
```diff
@@ -203,4 +258,125 @@ function besselk_large_orders_scaled(v, x::T) where T <: Union{Float32, Float64,
...
+No checks are performed on nu so this is not accurate when nu is an integer.
+"""
+function besselk_power_series(v, x::T) where T
+    MaxIter = 1000
+    z  = x / 2
+    zz = z * z
+    
+    logz = log(z)
+    xd2_v = exp(v*logz)
+    xd2_nv = inv(xd2_v)
```
heltonmc, line 356: I'm still unsure if this confuses what is going on here and shouldn't just be the literal power. Considering I think the powers are going to much faster on 1.9....

cgeoga: Was actually thinking about this too. Maybe @oscardssmith should make a `BetterPow.jl` in the mean time, even, and we could then use that....

heltonmc: It is about 5-10% faster.... but inconsistent with the rest of the power series. Maybe we will leave it for now with a note to look later...

src/besseli.jl
```diff
@@ -130,26 +130,91 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+            return cispi(abs_nu) * besseli_positive_args(abs_nu, abs_x)
+        end
     else
-        return T(besseli_large_orders(nu, x))
+        if x >= 0
+            return besseli_positive_args(abs_nu, abs_x) + 2 / π * sinpi(abs_nu) * besselk_positive_args(abs_nu, abs_x)
+        else
+            Iv = besseli_positive_args(abs_nu, abs_x)
+            Kv = besselk_positive_args(abs_nu, abs_x)
+            return cispi(abs_nu) * Iv + 2 / π * sinpi(abs_nu) * (cispi(-abs_nu) * Kv - im * π * Iv)
```
heltonmc, line 184: We also could return complex arguments for real arguments. I'm not for sure if that is a desired effect.... we don't support complex arguments yet so I don't know if it's best to only allow this if you call complex arguments.

heltonmc:
I think this is ready to merge now. I think the last thing is still the idea of potentially returning complex numbers for real (negative) inputs. Mathematica will return complex values for real inputs but `SpecialFunctions.jl` will not.

```julia
julia> using SpecialFunctions

julia> bessely(1.2, -0.2)
ERROR: DomainError with -0.2:
`x` must be nonnegative.
Stacktrace:
 [1] bessely(nu::Float64, x::Float64)
   @ SpecialFunctions ~/.julia/packages/SpecialFunctions/NBIqR/src/bessel.jl:556
 [2] top-level scope
   @ REPL[2]:1

julia> bessely(1.2, -Complex(0.2))
3.8701259814738727 - 2.904048804819615im
```

Matlab will output complex numbers for real inputs.
```matlab
>> bessely(1.2, -0.2)

ans =

   3.8701 - 2.9040i

>> bessely(1.2, -complex(0.2))

ans =

   3.8701 - 2.9040i
```
I'm wondering what the best is here?

oscardssmith:
Julia's general practice is to throw a domain error in cases like this (and require the input to be complex)

heltonmc:
That sounds good to me. One issue is that we don't actually support arguments for nonzero imaginary parts. So we would probably have to have a additional checks for nonzero complex terms.

oscardssmith:
Honestly, I think we can not deal with this until we start on more general complex support.

heltonmc:
Ya that sounds good to me.... I'll make an issue about it just so people are aware and put it on the updated readme!

Issue #30, opened 2022-08-04
Could return complex values for real inputs if argument is negative


heltonmc:
If called with negative arguments the routines could return imaginary values.
```julia
julia> besselj(0.4, 2.5)
0.24183102113825858

julia> besselj(0.4, -2.5)
0.07472989529876903 + 0.22999496849585185im
```

It appears the Julia standard is to return a domain error here unless called with complex. SpecialFunctions.jl will return error unless explicitly defined.
```julia
julia> using SpecialFunctions
julia> besselj(0.4, 2.5)
0.241831021138255

julia> besselj(0.4, -2.5)
ERROR: DomainError with -2.5:
`x` must be nonnegative.
Stacktrace:
 [1] besselj(nu::Float64, x::Float64)
   @ SpecialFunctions ~/.julia/packages/SpecialFunctions/NBIqR/src/bessel.jl:488
 [2] top-level scope
   @ REPL[5]:1

julia> besselj(0.4, -Complex(2.5))
0.07472989529876822 + 0.22999496849584836im
```

However, we do not yet support arguments with non zero imaginary components.

Edit: the answer in higher precision is `0.07472989529876900979841 +0.2299949684958517462472im

PR #32, opened 2022-08-04
remove complex output for real inputs

heltonmc:
Fixes https://github.com/heltonmc/Bessels.jl/issues/30. Removes functionality for all negative arguments that could output complex numbers for real inputs. Instead we return error here until support for complex numbers is provided.


heltonmc:
Just some coverage error. I've left the support for negative arguments only if the return value is real which I think is the appropriate way to do it for now.

PR #38, opened 2022-08-08
Add spherical bessel functions

### LINKED ISSUE #34: Should we support Spherical bessel functions?
Please upvote/down-vote if you would like to see support for spherical bessel functions.

These are fairly simply defined, and SpecialFunctions.jl does offer them https://github.com/JuliaMath/SpecialFunctions.jl/pull/196

However, I'm suspicious of their implementation https://github.com/JuliaMath/SpecialFunctions.jl/blob/36c547b4a270b6089b1baf7bec05707e9fb8c7f9/src/bessel.jl#L700-L707 ...... and the cutoffs they chose.

For example,
```julia
julia> using SpecialFunctions
julia> sphericalbesselj(0.5, 1e-200)
0.0

julia> sphericalbesselj(0.5, big"1e-200")
6.266570686577501256039413212027613132517466851524845791574808940855734136519535e-101
```
This seems like a bug to me and can be done in Float64. For example in Bessels.jl using Float64...
```julia
julia> x = 1e-200
1.0e-200
julia> v = 0.5
0.5

julia> sqrt(pi/2x)*besselj(v+1/2, x)
6.266570686577501e-101
```

You actually get a correctly rounded answer....

heltonmc:
Based on feedback in #34 and at https://github.com/SpM-lab/SparseIR.jl/pull/33 it would be nice to provide fast versions of spherical bessel functions.

Closes #34. 

```julia
julia> @benchmark Bessels.sphericalbesselj(10, z) setup=(z = rand()*100)
 Range (min … max):  35.105 ns … 195.677 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     35.895 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   45.506 ns ±  29.933 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▃▂▁                              ▁   ▁   ▁                  ▁
  ████▅▃▁▁▃▁▁▁▄▃▃▁▇▃▁▄█▅▃▁▄▄█▄▄▅█▆▄▆█▆▄▄█▇▃▅█▇▁██▇▄██▇▅▁█▇▇▄█▆ █
  35.1 ns       Histogram: log(frequency) by time       169 ns <

julia> @benchmark SpecialFunctions.sphericalbesselj(10, z) setup=(z = rand()*100)
BenchmarkTools.Trial: 10000 samples with 825 evaluations.
 Range (min … max):  149.190 ns …   2.253 μs  ┊ GC (min … max): 0.00% … 58.84%
 Time  (median):     275.410 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   552.528 ns ± 311.294 ns  ┊ GC (mean ± σ):  0.07% ±  1.18%

          █▅                                                     
  ▃▃▄▃▂▁▁▁██▂▂▁▁▂▂▂▂▂▁▁▁▁▁▁▁▁▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▄▄▄▄▅▄▄▄▄▄▄▄▄ ▃
  149 ns           Histogram: frequency by time          960 ns <

```

I think there are still some things to go on this for now but at least for integer orders this is much faster.


heltonmc:
I've added better handling of `Float16` and just generic input types. Hoping to update a lot of functions this way. @oscardssmith Are you ok with adding this functionality and the implementation? It appears to be of interest...

PR #39, opened 2022-08-09
support for Airy functions

heltonmc:
I think we should also provide the Airy functions as they can be computed with the bessel functions (which is what the SpecialFunctions.jl) does using the connection formulas https://dlmf.nist.gov/9.6.

The issue with this is that for negative arguments there is some cancellation. I found that the SpecialFunctions.jl routines are very inaccurate so testing against them is tricky. Though, these implementations suffer from same cancellation.


oscardssmith:
You know way more about this than I do, but IMO, these definitions belong in SpecialFunctions.jl and the way to improve them is to make SpecialFunctions use Bessels.jl

heltonmc:
I'm interested to see comments on the issue you posted in SpecialFunctions.jl..... In some sense there seems to be a lot of hesitancy (uncertainty?) from past threads about changing or moving code and that SpecialFunctions.jl should be somewhat stable and legacy (as it has a lot of dependents).

 I'm very in favor of having a special functions library that is entirely written in julia. It just makes it easier for all of us to diagnose and improve the code base. Honestly, I'm in favor of expanding this repository to house general special functions (struve, kelving etc) for julia implementations.. but I'm open with whatever people over there want to do.

Though, I think having a library of airy, hankel, and bessel functions in julia is very important and have to be the most used functions in SpecialFunctions.jl

For the airy functions. Let's look at one case.
```julia
julia> using SpecialFunctions

julia> airyaiprime(-55.71475517223996)
4.246454776939346e-5
```

The answer in high precision is `4.24645478021884647798774E-5` so that is barely 8 digits of accuracy for this... definitely makes it hard to test against.... I need to switch over to Arb for all these but testing a lot of numbers can take a really long time..

heltonmc:
Are you ok with merging this and #38 for the time being then? I'm hoping to also fix the NaN bug and release v0.2.0.

PR #43, opened 2022-08-15
Take Float32 a little more seriously

heltonmc:
Instead of casting all Float32 numbers to Float64 we can dispatch the cutoff for either `Float32` or `Float64`. I've also changed some of the structure so when we add BigFloat we can also just add in those cutoffs.

One thing I chose is to cast Float32 into Float64 within each algorithm and then do the calculation to Float32 precision. I think this is the better way to go as we usually get correctly rounded results.


src/U_polynomials.jl
```diff
@@ -36,12 +39,47 @@ function besseljy_debye(v, x)
...
+# Float128 - provide roughly ~1e-35 precision
+besseljy_debye_fit(x) = 16.0 + 1.0012*x + Base.Math._approx_cbrt(Float64(27.91*x))
+besseljy_debye_cutoff(nu, x) = nu > besseljy_debye_fit(x) && nu > 40
+
+#####
+##### Debye large order expansion coefficients
+#####
+
+function Uk_poly_Jn(p, v, p2, x::Float64)
+    if v > 5.0 + 1.00033*x + Base.Math._approx_cbrt(1427.61*x)
```
oscardssmith, line 67: It's probably better to move the multiplication outside the cbrt.

heltonmc: You mean like `5.0 + 1.00033*x +11.45 * Base.Math._approx_cbrt(x)`?

heltonmc:
These changes don't affect F64 performance while giving some performance benefits to using Float32
```julia
julia> @benchmark Bessels.bessely(v, x) setup=(v=Float32(rand()*100); x = Float32(rand()*100))
BenchmarkTools.Trial: 10000 samples with 980 evaluations.
 Range (min … max):   61.862 ns … 476.701 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     105.570 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   157.837 ns ± 100.683 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇█      ▇▃             ▆▂       ▁▁▁▁▁▂▂▂▂▂▂▂▂▃▃▃▃▂▂▂▂▂▂▂▁▁▁   ▂
  ██▇▄▄▂▃▂██▅▄▃▃▄▄▅▆▇▇▆▇▇██▆▆▅▄▄▇▇█████████████████████████████ █
  61.9 ns       Histogram: log(frequency) by time        367 ns <

julia> @benchmark Bessels.bessely(v, x) setup=(v=Float64(rand()*100); x = Float64(rand()*100))
BenchmarkTools.Trial: 10000 samples with 970 evaluations.
 Range (min … max):   76.632 ns … 562.328 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     147.165 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   229.631 ns ± 139.823 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █  ▆▁     █▂ ▁▅▂           ▁  ▁▁▁  ▁▁▁▁▁▁▁▂▃▂▃▄▃▃▃▃▃▃▃▃▃▃▂▂▁  ▂
  █▆▃██▅▄▂▅▅██████▆▆▅▂▄▄▅▅▆▇███████████████████████████████████ █
  76.6 ns       Histogram: log(frequency) by time        480 ns <

```

PR #47, opened 2022-08-19
Update sphericalbesselk and add sphericalbesseli

### LINKED ISSUE #25: Half-integer order for `besselk`

heltonmc:
See my comment at https://github.com/JuliaMath/Bessels.jl/pull/46#issuecomment-1220877330 for details.

But this updates some things from #46 and should close #25


heltonmc:
I probably should add the modified spherical bessel function of first kind here too. However, `besseli` is fast in general and forward recurrence is always unstable so there really isn't much to optimize for here like there is for other spherical bessel functions.

Either way it might be best to add specialized routines for `i0, i1, i2, i3, i4` and just use `besseli` for everything else.

src/besseli.jl
```diff
@@ -170,6 +170,10 @@ _besseli(nu, x::Float16) = Float16(_besseli(nu, Float32(x)))
 
 function _besseli(nu, x::T) where T <: Union{Float32, Float64}
     isinteger(nu) && return _besseli(Int(nu), x)
+    if ~isfinite(x)
+        isnan(x) && return x
```
oscardssmith, line 174: just `return x` unconditionally?

heltonmc:
Need to rebase considering changes from #48 and merge. Also need to increment version here.

PR #49, opened 2022-09-01
Create Invalidations.yml

ranocha:
This is based on https://github.com/julia-actions/julia-invalidations. Adding such checks came up in https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359. I suggest to add this check here since this package is widely used as a dependency.

See also SciML/MuladdMacro.jl#26 and SciML/MuladdMacro.jl#29

### DIFF
--- a/.github/workflows/Invalidations.yml
+++ b/.github/workflows/Invalidations.yml
@@ -1,0 +1,40 @@
+name: Invalidations
+
+on:
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: always.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  evaluate:
+    # Only run on PRs to the default branch.
+    # In the PR trigger above branches can be specified only explicitly whereas this check should work for master, main, or any other default branch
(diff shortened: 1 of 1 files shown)


oscardssmith:
I'm pretty sure this package isn't a widely used dependency (currently 0 dependents according to JuliaHub), but this is still good to have (especially because it will hopefully soonish be used by SpecialFunctions.jl.

ranocha:
Ah, sorry, I just copied the text over and though Bessel functions should be used widely :sweat_smile:

oscardssmith (approved):

heltonmc:
Hopefully it will be more used in the future!! Thank you for contributing this.

And just so I'm aware, when it fails you then have to manually look to see what is causing it? Though, hopefully PRs will be smallish from now on so easier to track down.

ranocha:
Yes, exactly. The number of invalidations on the default branch and the PR will be reported in CI but you need to dig manually to see what's going on. But - in my experience - fixing invalidations is quite a manual process anyways, so CI can't help more than telling you "hey, watch out".

heltonmc:
Great! Thank you for taking the time to do this for the ecosystem.

PR #51, opened 2022-09-24
Airy function in entire complex plane

heltonmc:
This PR gives dedicated routines for the airy functions instead of relying on the relations to Bessel functions. The first thing this allows is significantly more accuracy than we had before in whole domain (for airybi these are roughly 4 more digits accurate). I've updated the tests to reflect that.

The second thing is they are much faster now.
```julia
julia> @benchmark SpecialFunctions.airybi(z) setup=(z = 4.0)
 Range (min … max):  1.257 μs …  2.107 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.267 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.271 μs ± 27.456 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▂▄▄█▇▇▅▃▁                                              
  ▁▂▄▅██████████▆▅▄▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁ ▂
  1.26 μs        Histogram: frequency by time        1.31 μs <

julia> @benchmark airybi(z) setup=(z = 4.0)
BenchmarkTools.Trial: 10000 samples with 988 evaluations.
 Range (min … max):  46.224 ns … 79.481 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     47.749 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   48.534 ns ±  3.767 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▃  █▄                                                     ▁ ▁
  █▁▄██▄▄▅▇▇▇▆▅▆▅▆▆▆▇██▆▆▇▅▅▅▄▆▅▄▄▃▄▄▄▃▄▃▃▄▃▄▄▁▃▄▃▁▁▃▄▃▃▁▄▆▇█ █
  46.2 ns      Histogram: log(frequency) by time      69.3 ns <

```

So having dedicated routines allows for much better accuracy and speed. Of course, I have also added support for the entire complex plane. The motivation for this is some of the uniform asymptotic expansions for bessel functions require evalulation of the airy functions (https://dlmf.nist.gov/10.20) so this may come in handy as we look at those for complex plane.

Of course there are still some domains I'm not too happy with in the complex plane. I will add notes in the file on this. I chatted with @cgeoga and a potential use here is to use the hypergeometric functions or the exponentially improved functions to try and bridge domains. We can piece that together better in the future.

I still have a couple of things to finish up on this. Mainly tidying up and adding docs to this but also getting rid of some of those constant gamma calls. There is a lot of code overlap that could be simplified especially in the asymptotic expansions. However, what we want to overlap with will depend if we want to compute both airyai and airyaiprime or airyai and airybi. Just depends. It might be something we revisit in the future. Looking at the asymptotic expansions we will want airyai and airyprime so it might best to combine those....

Need to also go through and think more carefully about type promotions. I don't think we need to actually promote to float64 in a lot of the algorithms this is just left over from things I was trying.


heltonmc:
Ok I've updated the docs and added the appropriate limits for infinite arguments therefore no error is return they will simply return the branches defined by the exponential or 1 / z.

There's one thing I'm still thinking about is how do we want to appropriately promote types in some of the series. For example the besseli series can always be computed without loss of precision but only for real arguments. For complex argument we will need to promote ComplexF32 to ComplexF64 and same happens here for the airy functions (`airybi`) which can be computed stably for real arguments but not so for imaginary arguments. Need to come up with a good structure to promote `Float32` for these use cases.

heltonmc:
Ok I think this is ready to go. I've added tests for Float32 arguments. This function oscillates around the zero axis for negative arguments so it is just easy to promote these to Float64. I'm open to other ideas the best way to do this but went with a structure like.
```julia
function airyai_power_series(x::ComplexOrReal{T}; tol=eps(T)) where T
    S = eltype(x)
    ...
   ...
    for i in 0:MaxIter
        ...
        abs(t) < tol * abs(ai1) && break
       ...
    end
    return ...
end
airyai_power_series(x::Float32) = Float32(airyai_power_series(Float64(x), tol=eps(Float32)))
airyai_power_series(x::ComplexF32) = ComplexF32(airyai_power_series(ComplexF64(x), tol=eps(Float32)))
```
Here is a plot of the relative errors in the entire complex plane compared to the special functions implementation.
airyai_error

heltonmc:
@oscardssmith Ok with merging?

src/airy.jl
```diff
@@ -1,107 +1,523 @@
...
+        # use computation for real numbers then convert to input type for stability
+        # for imag(z) != 0 use rotation identity (http://dlmf.nist.gov/9.2.E14)
+        if iszero(y)
+            xabs = abs(x)
+            xx = 2 * xabs^(T(3)/2) / 3
+            Jv, Yv = besseljy_positive_args(one(T)/3, xx)
+            Jmv = (Jv - sqrt(T(3)) * Yv) / 2
+            return convert(eltype(z), sqrt(xabs) * (Jmv + Jv) / 3)
+        else
+            return exp(pi*im/3) * _airyai(-z*exp(pi*im/3))  + exp(-pi*im/3) * _airyai(-z*exp(-pi*im/3))
```
oscardssmith, line 64: I don't trust the `exp(pi*im/3)` to constant fold (especially pre 1.8). `cispi(1/3)` is probably a bit better.

src/airy.jl
```diff
@@ -1,107 +1,523 @@
...
     if x > zero(T)
-        return isinf(z) ? zero(T) : -x * besselk(T(2)/3, z) / (sqrt(T(3)) * π)
-    elseif x < zero(T)
-        Jv, Yv = besseljy_positive_args(T(2)/3, z)
-        Jmv = -(Jv + sqrt(T(3))*Yv) / 2
-        return isinf(z) ? throw(DomainError(x, "airyaiprime(x) is not defined.")) : x_abs * (Jv - Jmv) / 3
-    elseif iszero(x)
-        return T(-0.2588194037928068)
+        # use relation to besselk (http://dlmf.nist.gov/9.6.E2)
+        zz = 2 * z^(T(3)/2) / 3
```
oscardssmith, line 95: how does `z^(T(3)/2)` compare to `sqrt(z^3)`? I'd expect the latter to be faster (if slightly less accurate)

src/airy.jl
heltonmc: This is a good catch! Thank you

src/airy.jl
oscardssmith: `sqrt(z)^3`is also worth checking as it prevents overflow.

oscardssmith: Also worth considering is `z*sqrt(z)`

heltonmc: O yes  - I like that much better. Will update. Hard to tell much about speed because this is within a pretty complex branch but preventing overflow is good here.

PR #53, opened 2022-10-03
Return sequence of Bessel functions

heltonmc:
Here's a first draft for https://github.com/JuliaMath/Bessels.jl/issues/52. I still need to think about this a bit more but these will return right answers....

I think things to discuss:
1. Supporting negative indexes (I've returned an error here as the Amos routine only supports positive orders)
2. Handling overflow in forward recurrence and starting underflow  in downward recurrence.

I think for 2 we may want to have a check that if the start value is zero we just error here instead of iterating through. That could create a very silent performance hit but want to think about that a little bit....

### DIFF
--- a/src/besseli.jl
+++ b/src/besseli.jl
@@ -164,11 +164,11 @@ end
 
 Modified Bessel function of the second kind of order nu, ``I_{nu}(x)``.
 """
-besseli(nu::Real, x::Real) = _besseli(nu, float(x))
+besseli(nu, x::Real) = _besseli(nu, float(x))
 
 _besseli(nu, x::Float16) = Float16(_besseli(nu, Float32(x)))
 
-function _besseli(nu, x::T) where T <: Union{Float32, Float64}
+function _besseli(nu::T, x::T) where T <: Union{Float32, Float64}
     isinteger(nu) && return _besseli(Int(nu), x)
     ~isfinite(x) && return x
     abs_nu = abs(nu)
@@ -203,6 +203,13 @@ function _besseli(nu::Integer, x::T) where T <: Union{Float32, Float64}
     else
         return sg * besseli_positive_args(abs_nu, abs_x)
     end
+end
+
+function _besseli(nu::AbstractRange, x::T) where T
+    (nu[1] >= 0 && step(nu) == 1) || throw(ArgumentError("nu must be >= 0 with step(nu)=1"))
+    out = Vector{T}(undef, length(nu))
+    out[end-1], out[end] = _besseli(nu[end-1], x), _besseli(nu[end], x)
+    return besselk_down_recurrence!(out, x, nu) 
 end
 
 """
--- a/src/besselj.jl
+++ b/src/besselj.jl
@@ -206,11 +206,11 @@ end
 Bessel function of the first kind of order nu, ``J_{nu}(x)``.
 nu and x must be real where nu and x can be positive or negative.
 """
-besselj(nu::Real, x::Real) = _besselj(nu, float(x))
+besselj(nu, x::Real) = _besselj(nu, float(x))
 
 _besselj(nu, x::Float16) = Float16(_besselj(nu, Float32(x)))
 
-function _besselj(nu, x::T) where T <: Union{Float32, Float64}
+function _besselj(nu::T, x::T) where T <: Union{Float32, Float64}
     isinteger(nu) && return _besselj(Int(nu), x)
     abs_nu = abs(nu)
     abs_x = abs(x)
@@ -252,6 +252,18 @@ function _besselj(nu::Integer, x::T) where T <: Union{Float32, Float64}
             spi, cpi = sincospi(abs_nu)
             return T((cpi*Jnu - spi*Ynu) * sg)
         end
+    end
+end
+
+function _besselj(nu::AbstractRange, x::T) where T
+    (nu[1] >= 0 && step(nu) == 1) || throw(ArgumentError("nu must be >= 0 with step(nu)=1"))
+    out = Vector{T}(undef, length(nu))
+    if nu[end] < x
+        out[1], out[2] = _besselj(nu[1], x), _besselj(nu[2], x)
+        return besselj_up_recurrence!(out, x, nu)
+    else
+        out[end-1], out[end] = _besselj(nu[end-1], x), _besselj(nu[end], x)
+        return besselj_down_recurrence!(out, x, nu)
     end
 end
 
--- a/src/recurrence.jl
+++ b/src/recurrence.jl
@@ -2,7 +2,7 @@
 # outputs both (bessel(x, nu_end), bessel(x, nu_end+1)
 # x = 0.1; k0 = besselk(0,x); k1 = besselk(1,x);
 # besselk_up_recurrence(x, k1, k0, 1, 5) will give besselk(5, x)
-@inline function besselk_up_recurrence(x, jnu, jnum1, nu_start, nu_end)
+function besselk_up_recurrence(x, jnu, jnum1, nu_start, nu_end)
     x2 = 2 / x
     while nu_start < nu_end + 0.5 # avoid inexact floating points when nu isa float
         jnum1, jnu = jnu, muladd(nu_start*x2, jnu, jnum1)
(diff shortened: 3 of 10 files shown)


heltonmc:
I've added this functionality for all the functions now `besselj`, `bessely`, `besselk`, `besseli` and the `hankel` functions. 

I'm still unsure the best approach if the starting values of the recurrence underflow as I can see advantages either way. Presumably you are using this method because it is efficient, so if we throw an error it will alert the user that this approach may not be efficient and they can adjust their range of `nu` or just use broadcasting for large orders. Alternatively, we could loop through until it is finite but that to my knowledge would require evaluation of each value to check if nonzero which would be quite inefficient.

An example would be something like `besselj(0:1000, 5.0)`. This function returns `0` in `Float64` precision at `besselj(210, 5.0)` so starting downward recurrence at 1000 would require almost 800 individual evaluations of `besselj` to reach a point where we can start recurrence.

@lucifer1004 @krcools Do you have a preference?

heltonmc:
No worries I appreciate the thoughts and feedback as I keep going back and forth on this! Matlab will return the correct results (zeros if underflow in that region and values for lower orders). It does look like there is some performance hit but it is able to account for this fairly well.

The plan for me forward is to have a good underflow check that is much faster than computing the `besselj` function and will work for most cases where underflow will occur. This will reduce the need to compute `besselj` many times while always returning the correct result. I'm working on finding as simple of a check for this now so hopefully that is the last piece.

Thank you!

src/besselj.jl
```diff
@@ -255,6 +255,40 @@ function _besselj(nu::Integer, x::T) where T <: Union{Float32, Float64}
...
+        end
+        if k > 1
+            out[k] = _besselj(nu[k], x)
+            out[1:k+1] = besselj_down_recurrence!(out[1:k+1], x, nu[1:k+1])
+            return out
+        else
+            return out
+        end
+    end
+end
```
heltonmc, line 288: @oscardssmith Could you look at these lines? I think I'm creating unnecessary slices here. And for some reason this PR is also increasing allocations during broadcasting...
```julia
# MASTER
julia> @benchmark besselj.(1:200, 100.0)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  43.333 μs … 66.209 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     43.542 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   44.040 μs ±  1.517 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇█▅▃▂▂▁▁   ▁▄▄▃▂▁    ▁                                      ▂
  ██████████████████▇▇▆██▇███▇▇▆▆▆▅▅▆▇▆▅▃▅▁▄▁▅▄▄▆▄▁▄▄▃▅▆▄▅▅▅▄ █
  43.3 μs      Histogram: log(frequency) by time      51.1 μs <

 Memory estimate: 1.77 KiB, allocs estimate: 1.

# THIS PR
julia> @benchmark besselj.(1:200, 100.0)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  60.916 μs …  1.395 ms  ┊ GC (min … max): 0.00% … 95.40%
 Time  (median):     61.500 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   64.875 μs ± 33.526 μs  ┊ GC (mean ± σ):  0.97% ±  2.13%

  █▆▄▄▄▂▁                                                     ▁
  █████████▇▇▇▇▇▆▅▅▄▄▃▄▃▃▄▄▃▃▄▃▁▃▁▃▄▁▃▁▁▁▁▁▁▁▁▃▃▁▃▄▁▁▃▁▁▃▃▁▄▃ █
  60.9 μs      Histogram: log(frequency) by time       133 μs <

 Memory estimate: 25.31 KiB, allocs estimate: 1385.
```

heltonmc: Though this PR significantly reduces this time.
```julia
# This PR no broadcast
julia> @benchmark besselj(1:200, 100.0)
 Range (min … max):  1.379 μs … 75.442 μs  ┊ GC (min … max): 0.00% … 97.56%
 Time  (median):     1.462 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.555 μs ±  2.070 μs  ┊ GC (mean ± σ):  3.75% ±  2.76%

   ▃▃▅▇███▇▆▅▃▃▁                   ▁▂▂▂▂▁                    ▃
  ███████████████▆▆▃▁▃▁▁▁▃▄▃▁▁▁▁▅▆████████▇█▇▆▅▅▅▆▆▇██▇█▇▇▆▆ █
  1.38 μs      Histogram: log(frequency) by time     2.08 μs <

 Memory estimate: 3.66 KiB, allocs estimate: 10.
```

oscardssmith: `@views`?

oscardssmith: i.e
```if k > 1
            out[k] = _besselj(nu[k], x)
            @views besselj_down_recurrence!(out[1:k+1], x, nu[1:k+1])
            return out
```

heltonmc: `@views` did remove an allocation. Something funny is going on as my old session with Revise is showing 1 allocation where new sessions are showing 150 allocations. Ill have to dig into this deeper.....

src/recurrence.jl
```diff
@@ -56,3 +84,12 @@ end
     return jnup1, jnu
 end
 =#
+function besselk_down_recurrence!(out, x::T, nu_range) where T
+    x2 = 2 / x
+    k = length(nu_range) - 2
+    for nu in nu_range[end-1:-1:2]
```
oscardssmith, line 90: `begin+1`?

heltonmc: Yes that's better practice, thank you. I'm pretty sure this should work for any `AbstractRange` provided `step(nu) == 1`...

heltonmc:
Ok I think everything looks good now. This will properly handle starting at underflow.
```julia
julia> besselj(0:500, 20.0)
501-element Vector{Float64}:
  0.16702466434057608
  0.06683312417584697
 -0.1603413519229914
 -0.09890139456044525
  0.13067093355485782
  0.15116976798238838
 -0.05508604956366364
 -0.18422139772058657
 -0.07386892884074697
  0.125126254647989
  0.18648255802393707
  0.061356303375948065
 -0.11899062431039419
  ⋮
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
```
So we get appropriate results even when a lot of values are zero. The performance cost  here is fairly small...
```julia
julia> @btime besselj(0:500, 20.0)
  7.859 μs (2 allocations: 6.50 KiB)
```
So for 500 Bessel functions it just takes several microseconds to perform even when the majority of the values underflow. Of course if there is no overflow it will be faster...
```julia
julia> @btime besselj(0:500, 400.0)
  1.820 μs (2 allocations: 8.12 KiB)
```
So those are worst case scenarios (2 allocations have to check for underflow). The best case we get when x > nu.
```
julia> @btime besselj(0:500, 600.0)
  1.332 μs (1 allocation: 4.06 KiB)
```
So in that case only 1 allocation is needed and very fast.

Anyway I think this is ready to go now.

README.md
```diff
@@ -120,6 +120,46 @@ julia> besselk0(x)
...
+  443.328 ns (2 allocations: 1.75 KiB)
+```
+This function will allocate so it is recommended that you calculate the Bessel functions at the top level of your function outside any hot loop. For example,
+
+```julia
+function bessel_sequence(x, orders)
+    J_nu = besselj(orders, x)
+    out = zero(x)
+    for i in eachindex(J_nu)
+        out += sin(x*i)*J_nu[i]
```
oscardssmith, line 155: you can probably do better than calling `sin` repeatedly here.

src/besseli.jl
```diff
@@ -205,6 +206,34 @@ function _besseli(nu::Integer, x::T) where T <: Union{Float32, Float64}
...
+        else
+            inu = _besseli(nu[k], x)
+        end
+        out[k] = inu
+        k -= 1
+        k < 1 && break
+    end
+    if k > 1
+        out[k] = _besseli(nu[k], x)
+        out[1:k+1] = besselk_down_recurrence!(out[1:k+1], x, nu[1:k+1])
```
oscardssmith, line 228: why are these copies needed?

README.md
heltonmc: Ya this was just a (very bad) toy example to try and tell people to not call it inside a loop. Ill come up with something better.

src/besseli.jl
heltonmc: I was hoping that simply having
```
if k > 1
        out[k] = _besseli(nu[k], x)
        besselk_down_recurrence!(out[1:k+1], x, nu[1:k+1])
end
```
would work and do that operation in place but for some reason it was not making that copy appropriately with out explicitly having the `out[1:k+1]` statement....

heltonmc: Ok I think this works...
```julia
if k > 1
            out[k] = _besselj(nu[k], x)
            tmp = @views out[1:k+1]
            besselj_down_recurrence!(tmp, x, nu[1:k+1])
            return out
        else
```
Looks like that gets rid of that extra allocation

heltonmc: That seems a bit subtle what is going on there but.....
```julia
julia> @btime besselj(0:500, 20.0)
  7.620 μs (1 allocation: 4.06 KiB)
```
It does remove all of the unnecessary allocations. 

You think that is ok?

oscardssmith: much better.

oscardssmith: you could also write this with `@view` to remove a line, but this is good as is.

heltonmc: Perfect - thank you! Was a bit stumped on properly using `@views` but this is very fast now.

PR #54, opened 2022-10-14
add loggamma float64

heltonmc:
Bunch of things to still settle here but need this function for a couple routines and to avoid overflow.

Here's a plots of relative errors so far.
Screenshot from 2022-10-14 16-11-17

So there's a problem area right now that I need to work out. Here's a closer view
Screenshot from 2022-10-14 16-12-50

This algorithm uses recurrence to go between [2,3]. For args below 2 this leads to a bad subraction.. Essentially `loggamma(1.9) = loggamma(2.9) - log(1.9)` which get's pretty similar.
```julia
julia> loggamma(2.9)
0.6028696102493114

julia> log(1.9)
0.6418538861723947
```
Searching for a better way to handle that area. I'm also trying to improve that rational function. Remez giving me a little trouble but hoping to improve that approximation a little bit more.

### DIFF
--- a/src/gamma.jl
+++ b/src/gamma.jl
@@ -66,3 +66,45 @@ function small_gamma(x)
     q = evalpoly(x, Q)
     return z * p / q
 end
+
+function loggamma(x::Float64)
+    if x > 13
+        return large_loggamma(x)
+    else
+        return small_loggamma(x)
+    end
+end
+function small_loggamma(x::T) where T
+    sign = 1
+    z = one(T)
+    p = 0
+    u = x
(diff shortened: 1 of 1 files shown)


heltonmc:
Ok - I think I went with a Taylor expansion around x = 1 and then the issue at x = 2 can be handled stably using the expansion at x = 1. This brings the error down everywhere. There might be a better approach but I plan to do a minimax on that small interval (0.8<x<1.2) to improve those results further. Hopefully also decrease the number of terms. (and need to improve branches and conditions)

PR #57, opened 2022-10-17
Fix very large argument expansion in besselj0 and friends

heltonmc:
A fix for https://github.com/JuliaMath/Bessels.jl/issues/56.

This reduces the error over the whole range
Screen Shot 2022-10-16 at 10 44 52 PM

There might be a better way to do this but you have to reduce x and xn first then make the subtraction so used the formula cos(x + y) = cosx cosy - sinx siny. Definitely open to other ideas. We could have a different branch within the `cos_sum` function but then we hit that statement for all arguments instead of just big ones. Alternatively we can just improve the `cos_sum` function for large arguments....

### DIFF
--- a/src/besselj.jl
+++ b/src/besselj.jl
@@ -57,6 +57,13 @@ function _besselj0(x::Float64)
             q2 = (-1/8, 25/384, -1073/5120, 375733/229376)
             p = evalpoly(x2, p2)
             q = evalpoly(x2, q2)
+            if x > 1e15
+                a = SQ2OPI(T) * sqrt(xinv) * p
+                xn = muladd(xinv, q, -PIO4(T))
+                s_x, c_x = sincos(x)
+                s_xn, c_xn = sincos(xn)
+                return a * (c_x * c_xn - s_x * s_xn)
+            end
         end
 
         a = SQ2OPI(T) * sqrt(xinv) * p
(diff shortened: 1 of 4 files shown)


oscardssmith:
the faster way to fix this would be to do a little extra reduction in cos_sum, but honestly this world pretty well as is.

heltonmc:
Happy to do whatever you think is best. I think your right that ideally this could happen within cos_sum. I think a little extra compensation there is fine because that section of the routine is already like 2.5x faster than for x < 25. And as we showed in the benchmarks the runtime seems to be really dominated by the slowest branch anyway....

heltonmc:
I think I'll just merge as is. I think in the future if we want to take a look at this whole section to provide full compensation this can be changed. 

But https://github.com/JuliaMath/Bessels.jl/issues/22 covers these concerns and documents this as an open issue if someone wants to tackle it (I think right now I want to focus on complex support, etc.). We just need to keep in mind in the future that if we go the route of fully compensating we must also test for x > 1e20.

oscardssmith:
that makes sense. I have somewhat gotten lost in the rabbit-hole that Julia's `rem_pio2` isn't very good.

heltonmc:
No worries that is good to know! It's something I definitely want to come back to and keep in mind considering all the other discourse right now.

PR #60, opened 2022-10-21
Add mutating sequence

### LINKED ISSUE #59: mutating bessels! for sequence
First of all thanks for this nice package!

My FR:
I have a usecase were I repeatedly calculate bessels of a certain range for n. It would be nice if I would not have to alocate a vector a each iteration but instead could use an preallocated array and mutate it with some kind of `besselj!(out, n, x)`.
E.g.
```[julia]
result = Vector{Float64}(undef, 1000)
n = 0:999
for i = 1:10
   besselj!(result, n, i)
   do_stuf(result)
end
```

heltonmc:
Closes #59 

Need to go through and test benchmarks to make sure there are no regressions and need to add methods for all the other Bessel functions. Always amazed how expensive allocating memory is...

```julia
julia> @benchmark Bessels.besselj!(x, 1:10, 1.0) setup=(x=zeros(Float64, 10))
BenchmarkTools.Trial: 10000 samples with 741 evaluations.
 Range (min … max):  169.345 ns … 328.243 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     173.259 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   174.458 ns ±  11.158 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▁██▃  ▂▂▁                                               ▂
  ▄▃▅▇▆▆████▆████▇▆▅▅▄▅▅▅▅▁▃▃▄▅▄▃▄▁▄▃▃▁▃▄▃▁▁▁▅▃▄▃▃▄▁▃▄▃▄▁▄▄▄▅▆▅ █
  169 ns        Histogram: log(frequency) by time        199 ns <

julia> @benchmark Bessels.besselj(1:10, 1.0) setup=(x=zeros(Float64, 10))
BenchmarkTools.Trial: 10000 samples with 585 evaluations.
 Range (min … max):  203.162 ns …  1.300 μs  ┊ GC (min … max): 0.00% … 74.22%
 Time  (median):     204.342 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   211.169 ns ± 39.150 ns  ┊ GC (mean ± σ):  0.66% ±  3.32%

  █▁▂▁                   ▁▂▁ ▁                                 ▁
  ████▆▅▆▄▄▄▃▆▄▄▄▄▄▅▄▃▃▄▇██████▆▅▄▄▃▁▄▃▄▃▅▅▄▄▄▄▄▅▆▄▆▆▅▄▄▅▆▆▆▄▅ █
  203 ns        Histogram: log(frequency) by time       319 ns <

 Memory estimate: 144 bytes, allocs estimate: 1.
```

I guess we also need to decide if we want to export the new methods.... probably require v0.3.0 release...

### DIFF
--- a/src/besseli.jl
+++ b/src/besseli.jl
@@ -229,8 +229,8 @@ function _besseli!(out::DenseVector{T}, nu::AbstractRange, x::T) where T
     end
     if k > 1
         out[k] = _besseli(nu[k], x)
-        tmp = @view out[1:k+1]
-        out[1:k+1] = besselk_down_recurrence!(tmp, x, nu[1:k+1])
+        tmp = @view out[begin:k+1]
+        besselk_down_recurrence!(tmp, x, nu[begin:k+1])
         return out
     else
         return out
--- a/src/besseli.jl
+++ b/src/besseli.jl
@@ -169,6 +169,8 @@ besseli(nu, x::Real) = _besseli(nu, float(x))
 
 _besseli(nu::Union{Int16, Float16}, x::Union{Int16, Float16}) = Float16(_besseli(Float32(nu), Float32(x)))
 
+_besseli(nu::AbstractRange, x::T) where T = besseli!(zeros(T, length(nu)), nu, x)
+
 function _besseli(nu::T, x::T) where T <: Union{Float32, Float64}
     isinteger(nu) && return _besseli(Int(nu), x)
     ~isfinite(x) && return x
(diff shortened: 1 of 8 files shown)


heltonmc:
Ok I've added all the methods for `besselj!`, `bessely!`, `besselk!`, and `besseli!`. There was also a bad allocation in the modified Bessel functions that I cleaned up where they now all don't allocate.

I'm not for sure how I feel about exporting all these different types of methods. If we do we should probably also export `sphericalbesseli` and `sphericalbesselk` in next big release.... but IDK.

PR #62, opened 2022-10-24
minor `gamma` cleanups.

oscardssmith:
This should arguably should use `Base.@assume_effects :terminates_locally` but that only exists on 1.8+.

### DIFF
--- a/src/gamma.jl
+++ b/src/gamma.jl
@@ -49,3 +49,11 @@ function _gamma(x::Float64)
     q = evalpoly(x, Q)
     return z * p / q
 end
+
+
+function gamma(n::Integer)
+    n < 0 && throw(DomainError(n, "`n` must not be negative."))
+    n == 0 && return Inf*float(n)
+    n > 20 && return gamma(float(n))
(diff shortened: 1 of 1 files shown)


heltonmc:
Awesome! I'm wondering if we should go ahead and solve https://github.com/JuliaMath/Bessels.jl/issues/58 here.

Essentially, that issue can be solved by having some check if `x` is an integer and (x < 21) to just return the base factorial lookup table....  I'd honestly be fine just to have a different dispatch for it. So you would only hit that branch if you input an integer `nu` but I guess we could also check for it within the routines... but it seems way easier to do it here than update all of the functions. Something like

```julia
function gamma(x::Integer)
    if x < 21
        if x > 0
            return float(factorial(x - 1))
        elseif x < 0
            return DomainError...
        else
            return Float64(Inf)
        end
    else
        return gamma(Float64(x))
    end
end
```

heltonmc (approved):

heltonmc:
Great - I'll merge and test if this solves that issue.

PR #63, opened 2022-10-25
Small dispatch fix for gamma

### LINKED ISSUE #58: Performance degradation for very small arguments for besselj
I have augmented my [gist](https://gist.github.com/jwscook/4712f7120870f32378f00e69781e604a) to compare the speed of besselj in Bessels vs SpecialFunctions. Bessels is overall faster except for some orders for very small arguments.

[BesseljTime.pdf](https://github.com/JuliaMath/Bessels.jl/files/9804121/BesseljTime.pdf)

heltonmc:
~~Unfortunately we are promoting the argument to Float64 before hitting the gamma calls....~~

```julia
# This PR
julia> @benchmark Bessels.besselj(10, x) setup=(x=rand()*0.0000000001)
 Range (min … max):  26.144 ns … 52.565 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     27.193 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   27.259 ns ±  0.996 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                      ▂█▅                                      
  ▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▂▂▅███▆▃▄▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▂ ▂
  26.1 ns         Histogram: frequency by time          29 ns <

# Master
julia> @benchmark Bessels.besselj(10, x) setup=(x=rand()*0.0000000001)
BenchmarkTools.Trial: 10000 samples with 988 evaluations.
 Range (min … max):  45.090 ns … 98.066 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     46.202 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   46.266 ns ±  1.262 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                    █▂  █▂                                     
  ▂▂▂▂▂▂▂▂▁▁▂▂▂▂▂▂▄███▄███▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂ ▃
  45.1 ns         Histogram: frequency by time        48.2 ns <

```
~~So this does improve computational time now..... but I'm not for sure we should do much more than this for #58.... hoping our Float32 tests are robust enough to catch if this promotion is even needed.~~

Edit: Just fixed the dispatch calls to actually hit the gamma integer branch and fixed the proper return value at x = 0. This does nothing to improve #58 as that promotion was needed for accuracy.

### DIFF
--- a/src/besselj.jl
+++ b/src/besselj.jl
@@ -348,10 +348,10 @@ In general, this is most accurate for small arguments and when nu > x.
 function besselj_power_series(v, x::T) where T
     MaxIter = 3000
     S = promote_type(T, Float64)
-    v, x = S(v), S(x)
+    x = S(x)
 
     out = zero(S)
-    a = (x/2)^v / gamma(v + one(S))
+    a = (x/2)^v / gamma(v + 1)
     t2 = (x/2)^2
     for i in 0:MaxIter
         out += a
--- a/src/besselj.jl
+++ b/src/besselj.jl
@@ -348,10 +348,10 @@ In general, this is most accurate for small arguments and when nu > x.
 function besselj_power_series(v, x::T) where T
     MaxIter = 3000
     S = promote_type(T, Float64)
-    x = S(x)
+    v, x = S(v), S(x)
 
     out = zero(S)
-    a = (x/2)^v / gamma(v + 1)
+    a = (x/2)^v / gamma(v + one(S))
     t2 = (x/2)^2
     for i in 0:MaxIter
         out += a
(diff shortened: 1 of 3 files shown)


heltonmc:
Ya that is not going to work we need that promotion for single precision. The result for gamma at zero returns NaN instead of Inf for 0 and -Inf for -0.0.

PR #65, opened 2022-11-03
add Float32 gamma

oscardssmith:
From my testing this is roughly 3x faster than  SpecialFunctions. Maximum error is .511 ulp. and the speed comparison for various ranges is:

large negatives: 43ns vs 92ns for specialfunctions
small negatives: 21ns vs 54ns
small positives: 8ns vs 26 ns
large positives: 21ns vs 70ns

### DIFF
--- a/src/gamma.jl
+++ b/src/gamma.jl
@@ -1,6 +1,5 @@
 # Adapted from Cephes Mathematical Library (MIT license https://en.smath.com/view/CephesMathLibrary/license) by Stephen L. Moshier
-gamma(x::Float64) = _gamma(x)
-gamma(x::Float32) = Float32(_gamma(Float64(x)))
+gamma(x::Union{Float32, Float64}) = _gamma(x)
 
 function _gamma(x::Float64)
     T = Float64
@@ -48,6 +47,31 @@ function _gamma(x::Float64)
     p = evalpoly(x, P)
     q = evalpoly(x, Q)
     return z * p / q
+end
+
+function _gamma(_x::Float32)
+    isfinite(_x) || return _x
+    x = Float64(_x)
+    if _x < 0
+        s = sinpi(x)
+        s == 0 && throw(DomainError(_x, "NaN result for non-NaN input."))
+        x = 1 - x
+    end
+    if x < 5
+        z = 1.
+        while x>1
+            x -= 1
+            z *= x
+        end
+        num = evalpoly(x, (1.0, 0.41702538904450015, 0.24081703455575904, 0.04071509011391178, 0.015839573267537377))
+        den = x*evalpoly(x, (1.0, 0.9942411061082665, -0.17434932941689474, -0.13577921102050783, 0.03028452206514555))
+        res = z*num/den
+    else
+        x -= 1
+        w = evalpoly(inv(x), (2.506628299028453, 0.20888413086840676, 0.008736513049552962, -0.007022997182153692, 0.0006787969600290756))
+        res = @fastmath sqrt(x) * exp(log(x*1/ℯ) * x) * w
+    end
+    return Float32(_x<0 ? π / (s * res) : res)
 end
 
 function gamma(n::Integer)


heltonmc:
Thanks for doing this Oscar! Could you just add some tests at https://github.com/JuliaMath/Bessels.jl/blob/master/test/gamma_test.jl for the `Float32` case to make coverage happy then can merge whenever.

oscardssmith:
will do

oscardssmith:
Float16 version and tests added. The Float16 version is 13-18ns for negative inputs and 6-11ns for positive inputs (big inputs take more time).

heltonmc:
Awesome! This looks good, thank you. There is just that instance of `_gamma(x)` that needs to be fixed now which should pass the tests.

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+function gamma(_x::Float32)
+    x = Float64(_x)
+    if _x < 0
+        s = sinpi(x)
+        s == 0 && throw(DomainError(_x, "NaN result for non-NaN input."))
+        x = 1 - x
+    end
+    if x < 5
+        z = 1.
+        while x>1
```
heltonmc, line 59:
```suggestion
        z = 1.0
        while x > 1
```

src/gamma.jl
```diff
@@ -1,8 +1,5 @@
-# Adapted from Cephes Mathematical Library (MIT license https://en.smath.com/view/CephesMathLibrary/license) by Stephen L. Moshier
-gamma(x::Float64) = _gamma(x)
-gamma(x::Float32) = Float32(_gamma(Float64(x)))
-
-function _gamma(x::Float64)
+# Float64 version adapted from Cephes Mathematical Library (MIT license https://en.smath.com/view/CephesMathLibrary/license) by Stephen L. Moshier
+function gamma(x::Float64)
     T = Float64
     if x < 0
         (isinteger(x) || x==-Inf) && throw(DomainError(x, "NaN result for non-NaN input."))
```
heltonmc, line 5:
```suggestion
        (isinteger(x) || x == -Inf) && throw(DomainError(x, "NaN result for non-NaN input."))
```

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+    end
+    if x < 5
+        z = 1.
+        while x>1
+            x -= 1
+            z *= x
+        end
+        num = evalpoly(x, (1.0, 0.41702538904450015, 0.24081703455575904, 0.04071509011391178, 0.015839573267537377))
+        den = x*evalpoly(x, (1.0, 0.9942411061082665, -0.17434932941689474, -0.13577921102050783, 0.03028452206514555))
+        res = z*num/den
```
heltonmc, line 65:
```suggestion
        res = z * num / den
```

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+    else
+        x -= 1
+        w = evalpoly(inv(x), (2.506628299028453, 0.20888413086840676, 0.008736513049552962, -0.007022997182153692, 0.0006787969600290756))
+        res = @fastmath sqrt(x) * exp(log(x*1/ℯ) * x) * w
+    end
+    return Float32(_x<0 ? π / (s * res) : res)
```
heltonmc, line 71:
```suggestion
    return Float32(_x < 0 ? π / (s * res) : res)
```

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+end
+
+function gamma(_x::Float16)
+    x = Float32(_x)
+    if _x < 0
+        s = sinpi(x)
+        s == 0 && throw(DomainError(_x, "NaN result for non-NaN input."))
+        x = 1 - x
+    end
+    x>13 && return Float16(ifelse(_x>0, Inf32, 0f0))
```
heltonmc, line 81:
```suggestion
    x > 13 && return Float16(ifelse(_x > 0, Inf32, 0f0))
```

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+	z = 1f0
+	while x>1
```
heltonmc, line 83:
```suggestion
	while x > 1
```

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+		x -= 1
+		z *= x
+	end
+	num = evalpoly(x, (1.0f0, 0.4170254f0, 0.24081704f0, 0.04071509f0, 0.015839573f0))
+	den = x*evalpoly(x, (1.0f0, 0.9942411f0, -0.17434932f0, -0.13577922f0, 0.030284522f0))
+    return Float16(_x<0 ? Float32(π)*den/(s*z*num) : z*num/den)
```
heltonmc, line 89:
```suggestion
    return Float16(_x < 0 ? Float32(π)*den / (s*z*num) : z * num / den)
```

src/gamma.jl
```diff
@@ -50,6 +47,49 @@ function _gamma(x::Float64)
...
+end
+
+_gamma(x) = gamma(x) #easier than fixing this in other places.
```
heltonmc, line 92: I think https://github.com/JuliaMath/Bessels.jl/blob/fb17d4e0a11784145460e3dc46251b4d3a3885f2/src/gamma.jl#L7 is the only instance of `_gamma(x)` so might be best to just change that. I don't think anything in the bessel routines uses `_gamma`

heltonmc (commented):
Just some general spacing requests for code consistency.

oscardssmith:
@heltonmc all the test failures are now in `besseljy_positive_args` any idea if that's related to this PR?

src/gamma.jl
```diff
@@ -20,9 +17,13 @@ function _gamma(x::Float64)
         w = muladd(w, evalpoly(w, s), 1.0)
         # avoid overflow
         v = x ^ muladd(0.5, x, -0.25)
-        y = v * (v / exp(x))
+        res = SQ2PI(T) * v * (v / exp(x)) * w
 
-        return SQ2PI(T) * y * w
+        if _x < 0
+            return π / (res * s)
```
heltonmc, line 23: This seems to be the issue. We are dividing by `s` which is the polynomial coefficients.

heltonmc: Also, the `s` looks to be type unstable considering it is defined as a Float64 in the top half.

oscardssmith: ah. Variable names strike again.

oscardssmith:
Good news about these changes: they improve the accuracy of the Float64 gamma for negative numbers from millions of ulps to 7. The version of the reflection formula you were using had some serious issues due to 1-x not being exact.

oscardssmith:
everything works!!!!!!!!!!!!!!!!!!

heltonmc:
This looks great to me !!! Thanks for putting all this work in! Might be time to consider that separate gamma package with the loggamma function conversation. (my git skills haven't been good enough to pull that off yet).

I believe there is also an instance in the code base where we directly used the reflection formula... https://github.com/JuliaMath/Bessels.jl/blob/d9f15efbee1c22712e210b5ad6b248994d6ecda3/src/besselk.jl#L453-L459
I haven't checked if that is causing any errors there but might be good to reconsider that

heltonmc (approved):

PR #68, opened 2022-11-23
Complex support for Bessel functions I0(z), I1(z), J0(z), and J1(z)

heltonmc:
Here is a very rough draft of getting functionality in the entire complex plane for bessel functions of zero order (besselj0 and besseli0). I plan to start with the common functions (orders=0,1) for these functions. The good thing is when using complex numbers these have an easy relation to eachother so we can get both `besselj0` and `besseli0` with same algorithm....

Of course - polynomial approximation (using remez) is somewhat limited in the regime. So there's really two approaches that work well.
1. Asymptotic expansion works in entire complex plane ( for abs(z) > 20.0)
2. Power series works in entire complex plane (for abs(z) < 5.0) and if angle(z) < pi/5.0 but slowly converging..

That leaves a somewhat difficult area when 5.0 < abs(z) < 20.0. Though, if angle(z) < pi/5.0 we can use the power series with more terms and get good results. But if angle(z) > pi/5 the power series has large cancellation. I tried several different algorithms for this region but the challenge is that all the algorithms struggle close to the imaginary axis.

After some tinkering, I started using the recent AAA algorithm (https://www.chebfun.org/examples/approx/AAAApprox.html) which has some good results. Here is the relative error compared to Arb in the complex plane..

Screenshot from 2022-11-23 13-11-34

```julia
#this PR
julia> median(log10.(Float64.(abs.((1 .- b./a)))))
-15.653559774527022

julia> maximum(log10.(Float64.(abs.((1 .- b./a)))))
-13.368247642554348

#SpecialFunctions
julia> median(log10.(Float64.(abs.((1 .- s./a)))))
-15.643025367775405

julia> maximum(log10.(Float64.(abs.((1 .- s./a)))))
-12.90480864656037
```
So this actually has similar median errors and a noticeable better maximum error than SpecialFunctions.jl (which struggles on the imaginary axis).

Of course, near the boundaries of the rational approximation shows larger error. I'm still going to try and work this out a little bit more. Maybe break this up in two regions or consider a wider region with more points though it is still very accurate. Of course, the rational branch needs some optimization as it allocates when it doesn't have to.

But, of course I'm having to use matlab for this which is a pain. It is hard to get this working in higher precision and hard to test easily.... I'll continue trying to improve that region some more but probably only after the Thanksgiving.


heltonmc:
Also need to decide what to do with the Kelvin functions because once we have the complex versions of these they are also easily defined. Not sure why SpecialFunctions doesn't have these but maybe just have the user do this themselves?
```julia

function kelvinb0(z)
    be = besselj0(z*cispi(3/4))
    return real(be), imag(be)
end

```

Scipy has them defined https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.bei.html#scipy.special.bei  as well as mpmath https://mpmath.org/doc/current/functions/bessel.html#kelvin-functions. It looks like there is a slightly different use of notation as scipy just gives you zero order...

heltonmc:
I've mostly fixed the rational approximations now. In the complex plane we should take advantage of the fact that the maximal errors on an enclosed surface will occur on the boundaries (max modulus principle). So we must only minimize errors on the enclosed boundaries. To reproduce the coefficients using chebfun.
```
a = 4.9; b = 20.1;
a2 = pi/1.92; b2 = pi/5.5;
N = 5000;
Z2 = [b .* exp(i * ((a2-b2).*rand(N, 1) + b2)); a .* exp(i * ((a2-b2).*rand(N, 1) + b2)); ((b-a).*rand(N, 1) + a) .* exp(i * a2); ((b-a).*rand(N, 1) + a) .* exp(i * b2)];
F2 = double(sqrt(vpa(Z2)) .* besseli(0, vpa(Z2)) ./ exp(vpa(Z2)));
[r, pol, res, zer, z, f, w, errcev, wt] = aaa(F2, Z2, 'degree', 23, 'tol', 1e-35);

a = 5.0; b = 20.0;
a2 = pi/2; b2 = pi/5;
Z = ((b-a).*rand(N, 1) + a) .* exp(i * ((b2-a2).*rand(N, 1) + a2));
F = sqrt(vpa(Z)) .* besseli(0, vpa(Z)) ./ exp(vpa(Z));
err = 1.0 - (reval(Z, z, f, w) ./ sqrt(Z) .* exp(Z)) ./ besseli(0, Z);

hold on
histogram(log10(abs(err)))
```
The coefficients will be `f, z, w` that are used in the script. The issue is that you don't get the same coefficients always and also I used a trick to scale the rational approximation by its asymptotic limit. This is necessary to get good errors. I've also used a boundary slightly larger than what we need.

The resulting relative errors look like (for `besseli0`)
Screenshot from 2022-11-29 11-25-03

and for `besselj0`
Screenshot from 2022-11-29 11-29-34

This has same median error as SpecialFunctions.jl and better maximal error as shown above.

## Benchmarks

The advantage of course is the benchmarks. There is three branches for this so let's benchmark them separately.
```
# power series small arguments
julia> @benchmark besseli0(x) setup=(x=rand() + rand()*im + 1.0 + 1.0im)
 Range (min … max):  24.276 ns … 40.733 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     24.599 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   24.633 ns ±  0.443 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                        ▄██▅                   
  ▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▂▁▂▁▁▁▂▁▂▂▂▂▂▂▂▂▃▇████▇▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  24.3 ns         Histogram: frequency by time        24.8 ns <

# asymptotic expansion large arguments
julia> @benchmark besseli0(x) setup=(x=rand() + rand()*im + 20.0 + 20.0im)
BenchmarkTools.Trial: 10000 samples with 966 evaluations.
 Range (min … max):  79.645 ns …  1.013 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     81.588 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   90.100 ns ± 28.252 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆█▁                                                     ▁▂  ▁
  ██████▇▇█▇▆▆▆▅▄▅▅▄▅▂▃▅▃▂▄▃▄▅▆█▇▆▅▅▅▄▃▄▅▅▃█▅▄▃▄▅▃▆▆▄▄▄▅▇▇███ █
  79.6 ns      Histogram: log(frequency) by time       180 ns <

# complex rational approximation
julia> @benchmark besseli0(x) setup=(x=rand() + rand()*im + 6.0 + 8.0im)
BenchmarkTools.Trial: 10000 samples with 930 evaluations.
 Range (min … max):  107.923 ns … 228.156 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     109.424 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   110.165 ns ±   6.535 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▂▂▁▆█▇▇▇▆▄▁ ▁▁                                               ▂
  ▅███████████▇█████▇▇▅▅▄▅▄▄▅▄▃▄▃▃▁▃▁▃▄▃▄▃▃▁▃▁▁▃▁▁▁▄▁▁▁▃▁▃▁▁▃▁▃ █
  108 ns        Histogram: log(frequency) by time        122 ns <

# there's a sub branch using more terms in the power series
julia> @benchmark besseli0(x) setup=(x=rand() + rand()*im + 6.0 + 1.0im)
BenchmarkTools.Trial: 10000 samples with 990 evaluations.
 Range (min … max):  44.482 ns … 103.366 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     45.095 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   45.605 ns ±   4.672 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂██▇                                                         ▂
  ████▇▃▁▅▁▆▇▆▆▆▄▄▅▃▃▄▅▄▄▄▅▄▅▆▆▆▆▆▅▄▄▅▆▄▄▄▄▅▅▁▄▄▁▄▃▃▅▄▃▄▁▄▃▁▄▃ █
  44.5 ns       Histogram: log(frequency) by time      58.1 ns <

```

Comparing to `SpecialFunctions.jl`
```
julia> @benchmark SpecialFunctions.besseli(0, x) setup=(x=rand() + rand()*im + 10.0 + 10.0im)
BenchmarkTools.Trial: 10000 samples with 131 evaluations.
 Range (min … max):  730.351 ns …  1.147 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     750.401 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   752.536 ns ± 13.841 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                  ▃█▆▂        ▁▂                                
  ▁▂▅█▇▄▂▂▂▂▁▁▂▄▇███████▇▆▄▃▄▇███▆▆▅▅▅▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁ ▃
  730 ns          Histogram: frequency by time          786 ns <

julia> @benchmark SpecialFunctions.besseli(0, x) setup=(x=rand() + rand()*im + 25.0 + 15.0im)
BenchmarkTools.Trial: 10000 samples with 195 evaluations.
 Range (min … max):  514.185 ns …  1.245 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     540.728 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   544.827 ns ± 54.831 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▅█                                                         
  ▃▃▄██▃▂▂▂▂▂▂▂▂▂▁▂▂▂▁▁▂▂▁▁▂▁▂▁▂▁▁▁▁▁▂▁▁▂▁▁▂▂▂▂▂▁▂▂▁▁▂▁▂▁▂▁▁▂▂ ▂
  514 ns          Histogram: frequency by time          904 ns <

```
There is a pretty large range in `SpecialFunctions.jl` that you can hit but in general this will be around 8-12x faster.

heltonmc:
I've added support for first order now (I1(z) and J1(z)). The errors are similar to the zero order expansion. Here is the relative error for I1(z).
Screenshot from 2022-11-30 12-54-24

I think this is ready to go now. @oscardssmith let me know if you have any concerns. I will implement the modified versions (K0, K1, Y0, Y1) in a separate PR. I think we can then add the Kelvin function after that if we want. I also think the files are getting a bit too long. It might be best to reorganize into separate files with for the implementations of the 0 and 1 order and the generic orders nu.

src/besseli.jl
```diff
@@ -130,6 +130,238 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+    end
+    if imag(z) < 0.0
+        z = conj(z)
+        check = true
+    end
+
+    if abs(z) > 10.0
+        zinv = 1 / z
+        e = exp(z)
+        s = sqrt(2 * z * π)
```
oscardssmith, line 232: use `sinv=sqrt(zinv)/sqrt(2*pi)` to save a division.

src/besseli.jl
```diff
@@ -130,6 +130,238 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+        p2 = zinv*evalpoly(zinv*zinv, (0.375, 0.1025390625, 0.2775764465332031, 1.993531733751297, 27.248827311268542, 603.8440767050702, 19718.37591223663, 890297.8767070678, 5.310411010968523e7, 4.043620325107754e9, 3.827011346598606e11, 4.406481417852279e13, 6.065091351222699e15, 9.83388387659068e17))
+        r = e / s * (p - p2) - im * (p + p2) / (e * s)
+    elseif abs(z) < 5.2
+        r = z*evalpoly(z*z, (0.5, 0.0625, 0.0026041666666666665, 5.425347222222222e-5, 6.781684027777778e-7, 5.651403356481481e-9, 3.363930569334215e-11, 1.5017547184527747e-13, 5.214426105738801e-16, 1.4484516960385557e-18, 3.2919356728148996e-21, 6.234726653058522e-24, 9.991549123491221e-27, 1.3724655389411017e-29, 1.6338875463584545e-32, 1.7019661941233902e-35, 1.564307163716351e-38))
+    else
+        if angle(z) <= π / 4.4
+            zz = z*z
+            zz2 = zz*zz
```
oscardssmith, line 272: style nit, but I'd either go with `zzzz` or `z2` and `z4`

test/besseli_test.jl
```diff
@@ -65,6 +65,15 @@ i1x_32 = besseli1x.(Float32.(x32))
 @test i1x_64 ≈ i1x64_SpecialFunctions
 @test i1x_32 ≈ i1x32_SpecialFunctions
 
+# test complex implementation
+
+for x in [0.0, 0.01, 0.5, 1.0, 2.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 12.0, 14.0, 15.0, 17.0, 18.0, 20.0, 25.0, 50.0, 150.0], a in 0:pi/12:2pi
+    z = x*exp(im*a)
```
oscardssmith, line 71: `cis`

src/besseli.jl
```diff
@@ -130,6 +130,238 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+        r = s1 / (s2 * sqrt(z) * exp(-z))
+        end
+    end
+    check && (r = conj(r))
+    return r
+end
+
+function besseli0(z::ComplexF32)
+    z = ComplexF64(z)
+    check = false
```
oscardssmith, line 220: I'm pretty sure there's a better name than `check`

src/besseli.jl
```diff
@@ -130,6 +130,238 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+                 0.35489432819051814 + 0.8984965319188546im, 0.15105140195420463 - 0.7254114647779558im, 
+                 0.401057852147248 - 0.0014084695668244449im, 0.5404571564588408 - 0.1885021348797114im,
+                 0.407770977742567 - 0.00680290179701475im, 0.3994965367795384 - 0.0024661141779376626im, 
+                 0.8312271064319111 - 1.0050503936637845im, 0.40006191058465196 - 0.002248719754950748im, 
+                 0.4055260568863696 - 0.01689191195791546im, 5.210832430161148 + 1.7413278357835171im, 
+                 0.40364747476765517 - 0.003288380266379im, 0.40062798907705705 - 0.0018845654535970921im, 
+                 -0.39383201186515915 - 0.17987628384333437im, -2.053184226968296 + 1.6069066509951657im, 
+                 0.40617427977861376 - 0.005320233415147144im, 0.4034434765080089 - 0.006086635360116373im,
+                 -0.07645988937623527 + 1.981923553555365im, 1.8844134122620073 + 0.24179792442042025im
+                ).*w
```
oscardssmith, line 201: Would you get better accuracy by moving the `.*w` to compile time where you could do it in arbitrary precision?

src/besseli.jl
```diff
@@ -130,6 +130,238 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+        check = true
+    end
+
+    if abs(z) > 17.5
+        zinv = 1 / z
+        e = exp(z)
+        s = sqrt(2 * z * π)
+        p = evalpoly(zinv*zinv, (1.0, 0.0703125, 0.112152099609375, 0.5725014209747314, 6.074042001273483, 110.01714026924674, 3038.090510922384, 118838.42625678325, 6.252951493434797e6, 4.259392165047669e8, 3.646840080706556e10, 3.8335346613939443e12, 4.8540146868529006e14, 7.286857349377656e16))
+        p2 = zinv*evalpoly(zinv*zinv, (0.125, 0.0732421875, 0.22710800170898438, 1.7277275025844574, 24.380529699556064, 551.3358961220206, 18257.755474293175, 832859.3040162893, 5.0069589531988926e7, 3.8362551802304335e9, 3.6490108188498334e11, 4.218971570284097e13, 5.827244631566907e15, 9.47628809926011e17))
+        r = e / s * (p + p2) + im * (p - p2) / (e * s)
```
oscardssmith, line 153: you might be able to squeeze a little more accuracy by removing the `zinv` from `p2` and then combining `p` and `p2` with a `muladd`.

src/besseli.jl
```diff
@@ -149,8 +149,8 @@ function besseli0(z::ComplexF64)
         e = exp(z)
         sinv =sqrt(zinv)/sqrt(2*pi)
         p = evalpoly(zinv*zinv, (1.0, 0.0703125, 0.112152099609375, 0.5725014209747314, 6.074042001273483, 110.01714026924674, 3038.090510922384, 118838.42625678325, 6.252951493434797e6, 4.259392165047669e8, 3.646840080706556e10, 3.8335346613939443e12, 4.8540146868529006e14, 7.286857349377656e16))
-        p2 = zinv*evalpoly(zinv*zinv, (0.125, 0.0732421875, 0.22710800170898438, 1.7277275025844574, 24.380529699556064, 551.3358961220206, 18257.755474293175, 832859.3040162893, 5.0069589531988926e7, 3.8362551802304335e9, 3.6490108188498334e11, 4.218971570284097e13, 5.827244631566907e15, 9.47628809926011e17))
-        r = e * sinv * (p + p2) + im * (p - p2) * sinv / e
+        p2 = evalpoly(zinv*zinv, (0.125, 0.0732421875, 0.22710800170898438, 1.7277275025844574, 24.380529699556064, 551.3358961220206, 18257.755474293175, 832859.3040162893, 5.0069589531988926e7, 3.8362551802304335e9, 3.6490108188498334e11, 4.218971570284097e13, 5.827244631566907e15, 9.47628809926011e17))
+        r = zinv * (e * sinv * muladd(p, z, p2) + im * muladd(p, z, -p2) * sinv / e)
```
oscardssmith, line 153: `r = e * sinv * muladd(p2, zinv, p) + im * muladd(p2, -zinv, -p) * sinv / e`

oscardssmith:
The middle regions of high error aren't great. How is the error of other implementations?

heltonmc:
### Imaginary piece
Along and near the imaginary axis the imaginary component tends towards zero. 

Here is the plot of the imaginary component of this implementation compared to SpecialFunctions.jl along the imaginary axis (true value should be zero).
Screenshot from 2022-11-30 15-56-32
So this implementation is more accurate (has better peak error) than the SpecialFunctions.jl implementation.


### Real component

The issue is the real component oscillates around zero along this axis. Here is the plot of the real component of the function.

Screenshot from 2022-11-30 15-58-35

### relative error
And of course more interestingly it's relative errors compared to Arb.
Screenshot from 2022-11-30 16-02-01


So the median error is very similar between the two (SpecialFunctions.jl amos routine) and this PR with the max relative error being better with this implementation. So just in summary the relative error is a bit higher there (peaks around 1e-13) because the function is oscillating around zero. 

Does that sound ok?

oscardssmith:
Sounds good. I have a kind of hard time believing this is optimal, but I don't have any better ideas.

heltonmc:
We are definitely limited by the rational approximation only using double precision. The other option here is that along the imaginary axis this is essentially just the regular Bessel function. So we add a branch in `besseli0` that if the imaginary component is reasonably small then we can use our implementation of `besselj0` which expands about the zeros. Now the polynomial approximations along the real line will have some validity in the complex plane but usually just limited to small imaginary components. So there is a possibility that we can reuse our real function implementation for `besselj0` here (at the cost of a branch of course).

heltonmc:
I looked into my suggestion and it does work but because the polynomial approximation was optimized over a very small range on the real axis it has limited range in the imaginary axis. Essentially it is accurate if imag(z) < 0.1. Which takes care of the highest error.... we could focus a new implementation for `besselj0` for limited phase (angle(z) < pi/6) or something but I'm not for sure how much better we will be able to do honestly. 

Here is the updated relative errors when using the branch with `besselj0` (can see that very accurate strip on the imaginary axis now)...
Screenshot from 2022-11-30 17-47-34

Might be worth adding that branch?

oscardssmith:
it might be worth doing a Taylor series around the real axis to expand the convergence radius but it would need testing to see if it's worthwhile

heltonmc:
I'll look more closely at this. The issue is that we will need something between 5 < z < 20. So I would imagine that we would need a lot of different series to accommodate this. 

Here is a 20 term series around x= 6.0
```julia
f(x) = evalpoly(x - 6.0, (0.150645257250997, 0.2766838581275652, -0.0983796168027956, -0.0393674983001446, 0.009276407324868192, 0.0015513507450481006, 
 -0.0003059706348625948, -0.000030037971984537233, 5.22718360526724e-6, 3.5106184266832173e-7, -5.511322019111159e-8, 
 -2.7609585487363798e-9, 3.960776046407193e-10, 1.5645081640241804e-11, -2.0691628903926786e-12, -6.69779110834955e-14, 
 8.222170622153217e-15, 2.2435948383728534e-16, -2.570533958518674e-17, -6.041898146081074e-19, 6.490573541546584e-20))
```
Which seems to do ok 
```julia
julia> x = 6.0 + 2.0im
6.0 + 2.0im

julia> f(x)
0.7135646140123887 + 0.9219813154073662im

julia> besselj(0, x)
0.7135646140123892 + 0.9219813154073705im
```
May need a couple more terms. The issue is that it doesn't really have too wide validity. 
```
julia> x = 9.0 + 0.0im
9.0 + 0.0im

julia> besselj(0, x)
-0.09033361118287629 - 6.073569847713903e-17im

julia> f(x)
-0.09033361119245198 - 0.0im
```
So it looks like it works ok for abs(z) < 2.0 from the point of expansion using 20 terms..

Ill try to look more into it though.

heltonmc:
Used Taylor expansion for `besseli1` now similar to `besseli0`. Errors are comparable.
Screenshot from 2022-12-01 15-08-34

oscardssmith:
With these changes, can you use a simpler approximation for the disk now that you don't have to deal with the annoying region near the real axis?

oscardssmith:
Also for the rational approximation can you use `@fastmath` division which is a lot faster for complex numbers?

heltonmc:
Yes - Ideally now I can go back and redo the AAA rational approximations over a smaller region and use a less degree approximation! It is incredibly finicky though so will take me a bit to sort that out :sweat_smile: ....

heltonmc:
> Also for the rational approximation can you use `@fastmath` division which is a lot faster for complex numbers?

Ahh yes I didn't realize I just applied @fastmath to the for loop. That other division should also be able to use fastmath. Let me test. I should say I tested that loop with out the @fastmath flag and errors were identical. That makes sense considering we don't have to worry about under/over flow here as the range is so tight.

oscardssmith:
Yeah. I don't love our current complex division algorithm. It's made to ensure maximum likelyhood of not losing all the digits at the expense of speed and accuracy in the non-canceling cases.

heltonmc:
> I don't love our current complex division algorithm.

It was a performance gotcha for me at first and I saw the discourse post mentioning something similar. We will have to see about the generic complex arithmetic in a for loop. Here we are unrolling all the divisions so we don't pay that cost.

I've updated the new coefficients. Shrinking the region decreased a 25 degree approximation to a 16 degree so pretty significant. I decided to use a rectangular geometry for this approximation. It has some overlapping regions with other algorithms but works very well. Here is the relative errors now for `besseli1`
Screen Shot 2022-12-04 at 6 07 01 PM

```julia
# This PR
julia> out = Float64.([1 - besseli1(x + im*z) / ArbNumerics.besseli(1, ArbNumerics.ArbFloat(x) + im*ArbNumerics.ArbFloat(z)) for z in rand(500)*20.0, x in rand(500)*20.0]);

julia> maximum(abs.(out))
2.4538674030421287e-15

julia> median(abs.(out_sf))
1.5283991032546192e-16


# Compared to SpecialFunctions
julia> out_sf = Float64.([1 - SpecialFunctions.besseli(1.0, x + im*z) / ArbNumerics.besseli(1, ArbNumerics.ArbFloat(x) + im*ArbNumerics.ArbFloat(z)) for z in rand(500)*20.0, x in rand(500)*20.0]);

julia> median(abs.(out_sf))
1.5283991032546192e-16

julia> maximum(abs.(out_sf))
4.936224305972901e-13
```

So this implementation has much better maximum errors.

The matlab script to get the coefficients now are pretty simple.
```

Z2 = [(4.45 + ((17-3.2).*rand(500, 1) + 3.2) * i) (((14-4.45).*rand(500, 1) + 4.45) + 17.0*i) (14.0 + ((17-3.2).*rand(500, 1) + 3.2) * i)  (((14-4.45).*rand(500, 1) + 4.45) + 3.2*i)];

F2 = double(sqrt(vpa(Z2)) .* besseli(vpa(1.0), vpa(Z2)) ./ exp(vpa(Z2)));
[r, pol, res, zer, z, f, w, errcev, wt] = aaa(F2, Z2, 'degree', 16, 'tol', 1e-35);


Z = [((14-4.45).*rand(500, 1) + 4.45) + ((17-4).*rand(500, 1) + 4) * i]
err = 1.0 - (reval(Z, z, f, w) ./ sqrt(Z) .* exp(Z)) ./ besseli(1, Z);
histogram(log10(abs(err)))
```

oscardssmith:
this looks great!

heltonmc:
Thanks for the push to improve this 😄 Next steps are to get `besselk0`and probably should also do the airy functions like this as well. Just tedious...

src/besseli.jl
```diff
@@ -130,6 +130,209 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+        else
+            r = evalpoly(_z - 16.0, (-0.1748990739836292, -0.09039717566130419, 0.09027444873123035, 0.01312662593374557, -0.007667362695010893, -0.0005550708142218287, 0.0002571415573791134, 1.0850297108500103e-5, -4.565690471161262e-6, -1.2026225777846727e-7, 4.9959717576483296e-8, 8.48815248845687e-10, -3.701703923085197e-10, -4.101051708969331e-12, 1.980421966657433e-12, 1.4173962555705988e-14, -8.014281597026939e-15, -3.5742021762369616e-17, 2.540522616136639e-17, 6.485026844702886e-20, -6.482772100803784e-20, -7.615209126543415e-23, 1.3608987282597849e-22, 2.2067196209386653e-26, -2.3923906484884365e-25, 1.4153681120888267e-28, 3.5743243599678463e-28, -4.139824487468731e-31, -4.595455176935322e-31, 7.2929256888043685e-34, 5.138919311361309e-34))
+        end
+        r = conj(r) 
+    elseif abs(z) < 5.5
```
oscardssmith, line 174: Is this branch still worthwhile? It's only being used for a very small section of the inputs.

src/besseli.jl
```diff
@@ -130,6 +130,209 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+function besseli0(z::ComplexF64)
+    isconj = false
+    if real(z) < 0.0
+        z = -z
+    end
+    if imag(z) < 0.0
+        z = conj(z)
+        isconj = true
+    end
+    if abs(z) > 17.5
```
oscardssmith, line 146: you should probably use `abs2` here since it's 3x faster.

src/besseli.jl
```diff
@@ -130,6 +130,209 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+        end
+        r = conj(r) 
+    elseif abs(z) < 5.5
+        # use power series for I0
+        r = evalpoly(z*z, (1.0, 0.25, 0.015625, 0.00043402777777777775, 6.781684027777777e-6, 6.781684027777778e-8, 4.709502797067901e-10, 2.4028075495244395e-12, 9.385966990329842e-15, 2.896903392077112e-17, 7.242258480192779e-20, 1.4963343967340453e-22, 2.5978027721077174e-25, 3.842903509035085e-28, 4.9016626390753635e-31, 5.4462918211948485e-34, 5.318644356635594e-37))
+    else
+        if angle(z) <= π / 5.0
```
oscardssmith, line 178: `if imag(z) <= atan(pi/5) * x` will save some trig

src/besseli.jl
```diff
@@ -130,6 +130,209 @@ function besseli1x(x::T) where T <: Union{Float32, Float64}
...
+        e = exp(z)
+        sinv =sqrt(zinv)/sqrt(2*pi)
+        p = evalpoly(zinv*zinv, (1.0, 0.0703125, 0.112152099609375, 0.5725014209747314, 6.074042001273483, 110.01714026924674, 3038.090510922384, 118838.42625678325, 6.252951493434797e6, 4.259392165047669e8, 3.646840080706556e10, 3.8335346613939443e12, 4.8540146868529006e14, 7.286857349377656e16))
+        p2 = evalpoly(zinv*zinv, (0.125, 0.0732421875, 0.22710800170898438, 1.7277275025844574, 24.380529699556064, 551.3358961220206, 18257.755474293175, 832859.3040162893, 5.0069589531988926e7, 3.8362551802304335e9, 3.6490108188498334e11, 4.218971570284097e13, 5.827244631566907e15, 9.47628809926011e17))
+        r = e * sinv * muladd(p2, zinv, p) + im * muladd(p2, -zinv, p) * sinv / e
+    elseif real(z) < 4.5
+        # use taylor series around the roots (slight offset) of J0(z)
+        # use relation I0(z) = J0(im*z)
+        _z = imag(z) + abs(real(z))*im
+        if real(_z) < 2.2
```
oscardssmith, line 158: you should check that LLVM is smart enough to nest these branches in smart ways or generating a jump table.

src/besseli.jl
heltonmc: This is a good point now that the taylor/power series eats up a lot of this. I'll have to test as that corner piece between the taylor/rational/power series was the most tricky but I think this can branch can be avoided (or at least better implemented no)

src/besseli.jl
heltonmc: It's kinda a messy implementation 🤷‍♂️ I think there's a better way to express this. I'll work a little more on this!

src/besseli.jl
heltonmc: Yep - there was another `abs` I can get rid of too because of first quadrant. Though, I think `if imz <= tan(π/5) * rez` is the right formula.

src/besseli.jl
heltonmc: O ya it's much faster to do that instead.

heltonmc:
Added your suggestions which got some noticeable speed improvements.

```julia
# before
julia> @benchmark besseli0(x) setup=(x=ComplexF64(20*rand() + 20*rand()*im))
 Range (min … max):  19.283 ns … 127.539 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     86.387 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   67.706 ns ±  27.998 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                 ▁                                 █            
  ▃▁▁▁▁▁▁▅▆▂▂▁▁▁▁█▂▅▂▂▂▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▅█▃▂▂▂▁▁▃▅▃▅ ▂
  19.3 ns         Histogram: frequency by time          101 ns <

# after
julia> @benchmark besseli0(x) setup=(x=ComplexF64(20.0*rand() + 20.0*rand()*im))
 Range (min … max):  13.276 ns … 134.898 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     79.247 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   56.728 ns ±  28.968 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

          █                                             ▂       
  ▃▁▁▁▁▁▁▃█▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂██▃▂▄▅ ▂
  13.3 ns         Histogram: frequency by time         85.8 ns <

```

oscardssmith:
I think this is now ready to merge.

PR #69, opened 2022-12-02
Add a documentation page

heltonmc:
This adds the start of a hosted documentation page through Documenter.jl. In general, I think we could provide better documentation of this package.

Documenter usually gives me some fits so hopefully can figure out the deployment and things.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,26 @@
+name: Documentation
+
+on:
+  push:
+    branches:
+      - master # update to match your development branch (master, main, dev, trunk, ...)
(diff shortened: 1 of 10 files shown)


heltonmc:
I'm going to merge and see if the documentation builds appropriately and go from there. The deployment always gives me issues.

I think we should export some of the other function like `sphericalbesseli` and `sphericalbesselk`, `besselj!`, `bessely!`, `besselk!`, and `besseli!`. Perhaps we should just also commit to exporting `gamma` as well. Bring the `loggamma` PR over and just also have that officially here.... Anyway, would be a good thing for v0.3.0.

PR #77, opened 2023-02-08
Improved Airy function implementations for real arguments

heltonmc:
## Motivation

Of course, Airy functions are important on their own but perhaps equally important they are frequently used for uniform asymptotic expansions (e.g., [http://dlmf.nist.gov/10.20.E4](https://dlmf.nist.gov/10.20.E4)). So having these be as accurate and fast as possible will be important for the best implementations of other functions such as Bessel functions.

## Current implementation

The current implementation is fine in that it is general. Most expansions and power series are done in their loop forms which allows iteration over any number of terms. This in general is much slower.

Current performance:
```julia
julia> @benchmark Bessels.airyai(x) setup=(x=rand()*100)
 Range (min … max):   9.092 ns … 236.486 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     14.765 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   24.373 ns ±  32.491 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▇█▆▄▂▁▁                                        ▃▃▁          ▂
  ██████████▇▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇████▆█▇▆▅▇▅▆ █
  9.09 ns       Histogram: log(frequency) by time       162 ns <

julia> @benchmark Bessels.airyai(x) setup=(x=rand()*100 - 100.0)
 Range (min … max):    9.009 ns …   1.225 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     344.595 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   376.568 ns ± 155.786 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄▃                 █▃ █▄ ▇▃▂     ▃▅ ▁▄ ▄  ▃ ▂  ▁     ▁        ▂
  ██▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁██▇██▇████▃▃▁▁█████▇█▇▄█▇██▄██▇█▃▆█▄█▄▃█▆▇ █
  9.01 ns       Histogram: log(frequency) by time        906 ns <

```
It is probably best on these benchmarks to look at the mean and std along with the hi and lo times because there are so many branches. So for positive arguments we are between about 9 and 162 ns with an average of 25 ns and for negative arguments the situation is much worse between 9 and 900 ns with an average of around 375 ns....

This PR improves this performance pretty significantly.
```julia
julia> @benchmark Bessels.airyai(x) setup=(x=rand()*100)
 Range (min … max):   9.051 ns … 33.033 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     10.678 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   10.650 ns ±  0.406 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▂ ▁                                              ▂▁▆█ ▅▃  ▁
  ▃▆█▁█▆▁▄▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅████▁██▄ █
  9.05 ns      Histogram: log(frequency) by time      10.8 ns <

julia> @benchmark Bessels.airyai(x) setup=(x=rand()*100 - 100.0)
 Range (min … max):    7.633 ns … 156.031 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     112.654 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   103.273 ns ±  30.688 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                            █    
  ▄▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁██▄▃ ▂
  7.63 ns          Histogram: frequency by time          118 ns <

```

I'm going to post more details about the accuracy as well. These in general are more accurate as well for most arguments as everything is unrolled and I've used taylor series around each root.

The main source of error is now computing `exp(2 * sqrt(x)^3 / 3)`. The rest of the routine has ULP < 2.0 on the whole line but that exponential factor increases it significantly. 

Still need to do some work on this which I'll get to over the next few days


src/Airy/airy.jl
```diff
@@ -0,0 +1,299 @@
...
+    end
+end
+
+function airyaix(x::Float64)
+    if x >= 2.06
+        return airyaix_large_pos_arg(x)[1]
+    elseif x >= 0.0
+        # taylor expansion at x = 1.5 
+        b = evalpoly(x - 1.5, (0.07174949700810541, -0.09738201284230132, 0.05381212275607906, -0.012387253709224428, -0.0013886523923485612, 0.0017615621096121207, -0.00048234107659157565, 2.9849780287371904e-5, 1.853661597722781e-5, -6.0773111966738585e-6, 6.406078250357068e-7, 8.5642265292882e-8, -3.876060196303256e-8, 4.929943737019422e-9, 1.5110638652930308e-10, -1.493604112262068e-10, 2.148584715338907e-11, -2.6814055261032026e-13, -3.827831388762196e-13, 6.164805942828536e-14, -2.2166191076964464e-15, -6.912167850804561e-16, 1.2624054278515299e-16, -6.42973178916429e-18, -9.091593675774033e-19, 1.9432657516901093e-19, -1.198995513927753e-20, -8.798710894927164e-22, 2.33256140820231e-22, -1.6391332233394835e-23, -6.091803198418045e-25))
+        return exp(2 * (sqrt(x)^3 / 3)) * b
```
oscardssmith, line 25:
```suggestion
return exp((2/3) * x * sqrt(x)) * b
```

src/Airy/airy.jl
```diff
@@ -0,0 +1,299 @@
...
+        # negative numbers return complex arguments
+        throw(DomainError(x, "Complex result returned for real arguments. Use complex argument: airyaix(complex(x))"))
+    end
+end
+
+# airyaiprime & airyaiprimex
+
+function airyaiprime(x::Float64)
+    if x >= 2.06
+        return exp(-2 * (sqrt(x)^3 / 3)) * airyaix_large_pos_arg(x)[2]
```
oscardssmith, line 37: same here.

src/Airy/airy.jl
```diff
@@ -0,0 +1,299 @@
...
+    end
+end
+
+function airyaiprimex(x::Float64)
+    if x >= 2.06
+        return airyaix_large_pos_arg(x)[2]
+    elseif x >= 0.0
+        # taylor expansion at x = 1.5
+        b = evalpoly(x - 1.5, (-0.09738201284230132, 0.10762424551215811, -0.03716176112767328, -0.005554609569394245, 0.008807810548060603, -0.002894046459549454, 0.00020894846201160332, 0.00014829292781782247, -5.469580077006472e-5, 6.4060782503570685e-6, 9.42064918221702e-7, -4.651272235563907e-7, 6.408926858125249e-8, 2.1154894114102434e-9, -2.2404061683931022e-9, 3.437735544542251e-10, -4.558389394375444e-12, -6.8900964997719525e-12, 1.1713131291374217e-12, -4.4332382153928926e-14, -1.451555248668958e-14, 2.7772919412733657e-15, -1.4788383115077865e-16, -2.181982482185768e-17, 4.858164379225273e-18, -3.117388336212158e-19, -2.3756519416303343e-20, 6.531171942966468e-21, -4.753486347684502e-22, -1.8275409595254135e-23, 6.955638082337959e-24, -5.582291197792755e-25, -8.519287088295366e-27, 6.00631612185601e-27))
+        return exp(2 * (sqrt(x)^3 / 3)) * b
```
oscardssmith, line 53: and here.

src/Airy/airy.jl
```diff
@@ -0,0 +1,299 @@
...
+    sq_x3 = im*sqrt(abs(x3))
+    c1 = exp(2 * sq_x3 / 3) # loses some precision here
+    a = muladd(1 / sq_x3, -q, p) / c1
+    b = muladd(1 / sq_x3, q, p) * c1
+    c = muladd(1 / sq_x3, -q1, p1) / c1
+    d = muladd(1 / sq_x3, q1, p1) * c1
+
+    xsqrt = sqrt(sqrt(complex(x)))
+    ai = real((b + im*a) / (PIPOW3O2(T) * xsqrt))
+    aip = imag((c + im*d) * xsqrt  / PIPOW3O2(T))
```
oscardssmith, line 105: these seem overly complicated.

src/Airy/airy.jl
```diff
@@ -0,0 +1,299 @@
...
+    
+function airyai_large_neg_arg(x::T) where T <: Float64
+    x3 = x * x * x
+    p = evalpoly(1 / x3, (1.5707963267948966, 0.13124057851910487, 0.4584353787485384, 5.217255928936184, 123.97197893818594, 5038.313653002081, 312467.7049060495, 2.746439545069411e7, 3.2482560591146026e9, 4.97462635569055e11, 9.57732308323407e13))
+    q = evalpoly(1 / x3, (0.1636246173744684, 0.20141783231057064, 1.3848568733028765, 23.555289417250567, 745.2667344964557, 37835.063701047824, 2.8147130917899106e6, 2.8856687720069575e8, 3.8998976239149216e10, 6.718472897263214e12, 1.4370735281142392e15))
+    p1 = evalpoly(1 / x3, (-1.5707963267948966, 0.15510250188621483, 0.4982993247266722, 5.515384839161109, 129.24738229725767, 5209.103946324185, 321269.61208650155, 2.812618811215662e7, 3.3166403972012258e9, 5.0676100258903735e11, 9.738286496397669e13))
+    q1 = evalpoly(1 / x3, (0.22907446432425577, 0.22511404787652015, 1.4803642438754887, 24.70432792540913, 773.390007496322, 38999.21950723391, 2.8878225227454924e6, 2.950515261265541e8, 3.97712331943799e10, 6.837383921993536e12, 1.460066704564067e15))
+
+    sq_x3 = im*sqrt(abs(x3))
+    c1 = exp(2 * sq_x3 / 3) # loses some precision here
```
oscardssmith, line 97: `cis` of the real version.

src/Airy/airy.jl
heltonmc: Ughh yes thank you!... I wonder if that will help improve ULP here. (Scaled versions peak at around <1.9 ULP and regular version grow unbounded for large `x`.

src/Airy/airy.jl
heltonmc: Yes you are right. Because just on real-line can simplify to
```julia
ai = 2*real(a*im / (PIPOW3O2(T) * xsqrt))
aip = -2*real(c*im * xsqrt  / PIPOW3O2(T))
```
And can probably make some more simplifications since we only need real part

heltonmc:
Have some more work to go especially setting up the best way to test everything. Need to use Arb to get more correct results as the SpecialFunctions.jl implementation is pretty inaccurate.

Screen Shot 2023-02-09 at 3 58 00 PM

In general, these are accurate to < 1 ULP in the regions between asymptotic expansions (-9 < x < 2) which is much better than the SpecialFunctions.jl implementation. For large arguments these error grows proportional to `exp((2/3) * x * sqrt(x))`.

We can see this effect by looking at the error in the scaled version which avoids that calculation.

Screen Shot 2023-02-09 at 4 01 39 PM

The scaled version peak around 2 ULP. This also much better handles underflow/overflow.
```julia
julia> Bessels.airyai(104.0)
7.44875215829328e-309

julia> SpecialFunctions.airyai(104.0)
0.0

julia> ArbNumerics.airyai(ArbNumerics.ArbFloat(104.0))
7.448752158292226089088662383555e-309
```

To do: (1) improve some of the code structure, (2) decrease amount of terms in taylor series (3) add docs (4) fix tests

src/Airy/airy.jl
```diff
@@ -75,41 +75,43 @@ function airyaix_large_pos_arg(x::T) where T <: Float64
...
         xsqr = sqrt(x)
         zinv = 3 / (2 * x * xsqr)
         xsqr = sqrt(xsqr)
         p = evalpoly(zinv, (9.99999999999999995305e-1, 1.40264691163389668864e1, 7.05360906840444183113e1, 1.59756391350164413639e2, 1.68089224934630576269e2, 7.62796053615234516538e1, 1.20075952739645805542e1, 3.46538101525629032477e-1))
         q = evalpoly(zinv, (1.00000000000000000470e0, 1.40959135607834029598e1, 7.14778400825575695274e1, 1.64234692871529701831e2, 1.77318088145400459522e2, 8.45138970141474626562e1, 1.47562562584847203173e1, 5.67594532638770212846e-1))
         p1 = evalpoly(zinv, (1.00000000000000000550e0, 1.39470856980481566958e1, 6.99778599330103016170e1, 1.59317847137141783523e2, 1.71184781360976385540e2, 8.20584123476060982430e1, 1.47454670787755323881e1, 6.13759184814035759225e-1))
         q1 = evalpoly(zinv, (9.99999999999999994502e-1, 1.38498634758259442477e1, 6.86752304592780337944e1, 1.53206427475809220834e2, 1.58778084372838313640e2, 7.11727352147859965283e1, 1.11810297306158156705e1, 3.34203677749736953049e-1))
-        ai = inv_sqpi * p / (q * 2 * xsqr)
-        aip = -0.5 * inv_sqpi * xsqr * p1 / q1
+        ai = ONEOSQPI(T) * p / (q * 2 * xsqr)
```
oscardssmith, line 90: fold the `/2` into the constant.

src/Airy/airy.jl
```diff
@@ -75,41 +75,43 @@ function airyaix_large_pos_arg(x::T) where T <: Float64
...
     end
 end
 
 function airyai_large_neg_arg(x::T) where T <: Float64
-    x3 = x * x * x
-    p = evalpoly(1 / x3, (1.5707963267948966, 0.13124057851910487, 0.4584353787485384, 5.217255928936184, 123.97197893818594, 5038.313653002081, 312467.7049060495, 2.746439545069411e7, 3.2482560591146026e9, 4.97462635569055e11, 9.57732308323407e13))
-    q = evalpoly(1 / x3, (0.1636246173744684, 0.20141783231057064, 1.3848568733028765, 23.555289417250567, 745.2667344964557, 37835.063701047824, 2.8147130917899106e6, 2.8856687720069575e8, 3.8998976239149216e10, 6.718472897263214e12, 1.4370735281142392e15))
-    p1 = evalpoly(1 / x3, (-1.5707963267948966, 0.15510250188621483, 0.4982993247266722, 5.515384839161109, 129.24738229725767, 5209.103946324185, 321269.61208650155, 2.812618811215662e7, 3.3166403972012258e9, 5.0676100258903735e11, 9.738286496397669e13))
-    q1 = evalpoly(1 / x3, (0.22907446432425577, 0.22511404787652015, 1.4803642438754887, 24.70432792540913, 773.390007496322, 38999.21950723391, 2.8878225227454924e6, 2.950515261265541e8, 3.97712331943799e10, 6.837383921993536e12, 1.460066704564067e15))
+    invx3 = 1 / x^3
```
oscardssmith, line 97: be aware that `(1/x)^3` is more resistant to overflow (but you might not care since `1/(x^3)` is more accurate in the absence of overflow).

src/Airy/airy.jl
```diff
@@ -231,37 +233,42 @@ function airybiprimex(x::Float64)
...
+    xsqrx = xabs * xsqr
+    scale = cis(2 * x * xsqr / 3)
+    b = muladd(im / xsqrx, p1, p) * scale
+    d = muladd(im / xsqrx, q1, q) * scale
+
+    xsqr = sqrt(xsqr * 0.5)
+    cxsqr = complex(xsqr, xsqr) # sqrt(im*xsqr)
+
+    bi = 2*real(b / (cxsqr * PIPOW3O2(T)))
+    bip = -2*real(cxsqr * d / PIPOW3O2(T))
```
oscardssmith, line 270: `-2 * cxsqr / PIPOW3O2(T) * real(d)`?

src/Airy/airy.jl
heltonmc: This is a good thing to be aware of and my next step will be to check really large arguments. For negative arguments these slowly decay so this might be important. Surprisingly, SpecialFunctions.jl errors pretty quickly...

```julia
julia> SpecialFunctions.airyai(-1e7)
ERROR: AmosException with id 4: input argument magnitude too large, complete loss of accuracy by argument reduction.
Stacktrace:
 [1] _airy(z::ComplexF64, id::Int32, kode::Int32)
   @ SpecialFunctions ~/.julia/packages/SpecialFunctions/ZpCCa/src/bessel.jl:42
 [2] airyai
   @ ~/.julia/packages/SpecialFunctions/ZpCCa/src/bessel.jl:74 [inlined]
 [3] airyai(x::Float64)
   @ SpecialFunctions ~/.julia/packages/SpecialFunctions/ZpCCa/src/bessel.jl:173
 [4] top-level scope
   @ REPL[40]:1

julia> airyai(ArbFloat(-1e7))
0.005418514944210602418513854511213353963157226168027276649720555925636741540454

## current PR
julia> airyai(-1e7)
0.00541851889376531
```

We might need to do better but I'll work on trying to handle this regime more gracefully. Checking against Arb this shouldn't underflow....
```julia
julia> setworkingprecision(ArbFloat, 10000); setextrabits(5000);

julia> Float64(airyai(ArbFloat(-1e308)))
5.538960160131485e-78
```
I'll work on these branches some and see...

src/Airy/airy.jl
heltonmc: I think there's a potential for speed up here by unrolling the real parts (ignore the imaginary parts) but `b`, `d`, and `cxsqr` are complex. Since `cxsqr` has identical real and imaginary parts we can probably just simplify `bip = -2 / PIPOW3O2(T)  * (real(d) - imag(d)) * real(cxsqr)` but also change `cxsqr` to just be one part. I'm wondering if we can just completely unroll all these imaginary parts away. This section is also prone to cancellation.... Ill work on this too

heltonmc: Yep this simplifies down to (which avoids complex divisions)..
```julia
b = muladd(im / xsqrx, p1, p) * scale
d = muladd(im / xsqrx, q1, q) * scale

xsqr = sqrt(xsqr * 0.5)
bi = (real(b) + imag(b)) / (xsqr * PIPOW3O2(T))
bip = -2 / PIPOW3O2(T) * (real(d) - imag(d)) * xsqr
return bi, bip
```

src/Airy/airy.jl
oscardssmith: Specifically, it only matters when `x^3` overflows, in which case `(1/x)^3` might be a subnormal.

src/Airy/airy.jl
oscardssmith: you can pull out the `*.5` from the `sqrt` but other than that, this looks a lot better. Complex division really sucks.

heltonmc: Unrolling all of this. I haven't checked it it is faster but probably (if we do `sincos` instead). It becomes...
```julia
    xabs = abs(x)
    xsqr = sqrt(xabs)
    xsqrx = xabs * xsqr
    z = 2 * x * xsqr / 3
    b = p * (sin(z) + cos(z)) + p1 * (cos(z) - sin(z)) / xsqrx
    d = q * (cos(z) - sin(z)) - q1 * (cos(z) + sin(z)) / xsqrx

    xsqr = sqrt(xsqr * 0.5)
    bi = b / (xsqr * PIPOW3O2(T))
    bip = -2 / PIPOW3O2(T) * d * xsqr
    return bi, bip
```
The `cos(z) - sin(z)` is prone to cancellation though so need to work on that more and shouldn't use this directly.

oscardssmith: that's just `sqrt(2)*sin(x-pi/4)`. I really need to write you a good function for `sin` with quarter-pi offsets.

heltonmc: Yes! It pops up everywhere. I remember seeing this in openlibm https://github.com/JuliaMath/openlibm/blob/1d2c5e3bf52e050c994510ede91e0d2f0717173a/src/e_j0.c#L29-L37... which offered that work around.

heltonmc: Recent commit has avoided any complex arithmetic here. Though now we must compute (cos(x) +- sin(x)). I'm tempted to have a separate issue for computing the trig functions with quarter-pi offsets as they pop up in several functions. We can update these when we get a good functions. This implementation is fairly strictly superior to the current version (more accurate and much faster) so I don't think we have to wait for such an implementation before merging.

TODO: 1. Use similar approach for the airyai function to avoid complex functions.
              2. Need to have a separate infinite branch for negative arguments to avoid computing `sin` at negative infinity. Overall function goes to zero.
              3. Add docs.
              4. Define a better test structure for this function.

heltonmc:
Unfortunately very large negative arguments seem to lose all precision. At this point the asymptotic expansion can be simplified down to.
```julia

function airyai_big_neg_arg(x::T) where T
    x = abs(x)
    z = 2 * x * sqrt(x) / 3
    a = cos(T(π)/4 - z) - 5 * sin(T(π)/4 - z) / (48*z) 
    return a / (sqrt(T(π)) * sqrt(sqrt(x)))
end
```

```julia
julia> airyai_big_neg_arg(-1e20)
2.683723726309941e-6

julia> airyai_big_neg_arg(-BigFloat(1e20))
-5.352000451708926557753870608474960578511367628202425320476606952213081987517468e-06
```
SpecialFunctions.jl just errors here. Here are a couple of attempts to reduce it ...

```julia
julia> function airyai_big_neg_arg(x::T) where T
           x = abs(x)
           z = 2 * x * sqrt(x) / 3
           a = (-cos(2z)/(sin(z) - cos(z))) - 5 * (-cos(2x)/(sin(x) + cos(x))) / (48*z) 
           return a / (sqrt(2)*sqrt(T(π)) * sqrt(sqrt(x)))
       end
airyai_big_neg_arg (generic function with 1 method)

julia> airyai_big_neg_arg(-1e20)
-1.6114955560981463e-6

julia> function airyai_big_neg_arg(x::T) where T
           x = abs(x)
           z = 2 * x * sqrt(x) / 3
           a = cos(mod2pi(z) - T(π)/4) + 5 * sin(mod2pi(z) - T(π)/4) / (48*z) 
           return a / (sqrt(T(π)) * sqrt(sqrt(x)))
       end
airyai_big_neg_arg (generic function with 1 method)

julia> airyai_big_neg_arg(-1e20)
-1.6114955560981482e-6
```
Which give similar results... I'm imagining we would have to reduce `z` better before the quarter-pi shift though as that is introducing error here (as we are reducing after the rounding of previous 3/2 power)....

heltonmc:
Large negative arguments are going to need some more care. Here is a chart of the errors in ULP (using FunctionAccuracyTests.jl) for `airyai(x)` the bottom right plot is absolute error. Notice some are on a log scale. Current PR in red dots and SpecialFunctions.jl in black dots.
Screenshot from 2023-02-17 10-11-43

For positive arguments everything is ok. The increase is from the rounding in the x^1.5 term which follows the implementation from SpecialFunctions.jl. This also handles subnormal numbers until underflow around `x~105`. The top right plot shows this well where the SpecialFunctions.jl returns bad results (premature underflow). The routine is also accurate to less than 2 ULPS between `-10 < x < 2` where SpecialFunctions.jl is sometimes very inaccurate. (They use the power series over too wide a range).

The real problem is obviously negative arguments... which I'm not for sure exactly what to do here. SpecialFunctions.jl eventually returns an error around `x = -1e6`. The bottom right plot shows the steady increase in absolute error as we approach neg. infinity. SciPy return NaN in this range and Matlab decides to just return the wrong result. (The results that this PR match what Matlab returns for large negative arguments). 

The problem can really be reduced to...
```julia
julia> x = 1e30
1.0e30

julia> z = 2 * x * sqrt(x) / 3
6.666666666666667e44

julia> cos(z)
0.9988572973157203

julia> x = big"1e30"
1.0e+30

julia> z = 2 * x * sqrt(x) / 3
6.666666666666666666666666666666666666666666666666666666666666666666666666666646e+44

julia> cos(z)
-0.999201068754754087403977945244253575674850548675701791155692542429313401216291
```
So we will need to decide if we want to just return wrong results (matlab approach), return NaN (scipy approach), return error (SpecialFunctions.jl approach) or try to fix this issue.....

src/Airy/airy.jl
heltonmc: Because this is only used for the polynomial for large arguments we just return the first term in the polynomials which is independent of `x` for large values. This occurs way before `x^3` overflows so I don't think this will matter here (considering also my comments below about negative arguments)

heltonmc:
Fixed a few issues mainly with large arguments/infs/nan. I've also added explicit cutoffs for large negative arguments by throwing an error based on feedback from slack (@ParadaCarleton and & Aaron Denney) when all loss of relative error is met. I still will return the correct limits for infinite arguments if there is one but that could change.

Left to do:
1. Fix test structure (still unsure exactly best way)
2. Add docs.

This behavior is probably a breaking change so should be bumped to v0.3.

src/Airy/airy.jl
```diff
@@ -265,17 +292,16 @@ function airybi_large_neg_arg(x::T) where T <: Float64
...
+    # prone to error for large arguments
+    spc, smc = sincos(mod2pi(z) + π/4)
+    b = p1 * spc / xsqrx + p * smc
+    d = q * spc - q1 * smc / xsqrx
 
-    xsqr = sqrt(0.5) * sqrt(xsqr)
-    bi = b / (xsqr * PIPOW3O2(T))
-    bip = -2 / PIPOW3O2(T) * d * xsqr
+    xsqr = sqrt(xsqr)
+    bi = 2 * b / (xsqr * PIPOW3O2(T))
```
oscardssmith, line 303: separate out `2/ PIPOW3O2(T)` here as well.

heltonmc: Fixed.

heltonmc:
Sorry this took so long to get right but I think these are looking really solid now. Code coverage decrease if from complex routines no longer needing to handle real arguments. I need to update the complex routines completely (separate PR) to avoid all the complex divisions so will remove that branch then.

I'm going to look through it again but I've gone with a different test structure. This could perhaps be improved but I've generated and stored a ton of values over a wide range (-1e300 to 1e300) with a higher density for smaller arguments. These values were generated with Arb in high precision and rounded and stored as Float64. This allows for more strict testing of accuracy. I'm hoping this type of structure can be used for all the functions eventually as it is much faster and we get more accurate testing results.

heltonmc:
This is a refactor of the code to take advantage of https://github.com/JuliaMath/Bessels.jl/pull/79 (SIMD). With that PR these functions can be sped up by about ~3.5x. I will reorg this code in that PR to keep the implementations separate.

I got minimax polynomials and rational functions working here which drastically reduces the number of coefficients to be used (avoiding taylor series). I've reorganized the code to separate the positive and negative domain considering we will typically want the positive arguments in the Airy type expansions for the Bessel functions.

I don't particularly like how these are unpure and throw the domain error so internally we can use the pure functions considering we would have already checked for the domain. I'm still not for sure I entirely like throwing the domain error instead of returning `NaN` like SciPy does but I'm open to whatever here.

There's a lot of boiler plate code that I think I can reduce a little bit so will work on that and might try to work more on the negative arguments..... For `airybi` I was also able to get minimax rational functions to work however they could only be evaluated in BigFloat precision. Once you convert the coefficients to Float64 I was only able to get the ULP around 3-5 which I wasn't as happy with. I eventually decided to go with the Chebyshev polynomials on this domain as they could be evaluated stably within 1 ULP. This could be a region in the future that could be improved upon.

heltonmc:
Ok I think this is ready. Sample benchmarks comparing this PR, master, and SpecialFunctions.
```julia

# this PR
julia> @benchmark Bessels.airyai(x) setup=(x=rand()*15.0)
 Range (min … max):  5.541 ns … 53.250 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.791 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.457 ns ±  1.611 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▆▁                                         ▃▇█▃  ▂    ▁   ▂
  ███▅██▆▇▅▃▄▁▁▃▃▃▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁████▅▇█▇▇▆▇██▆ █
  5.54 ns      Histogram: log(frequency) by time     9.62 ns <

# master
julia> @benchmark Bessels.airyai(x) setup=(x=rand()*15.0)
 Range (min … max):   7.299 ns … 217.300 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     26.569 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   68.317 ns ±  59.197 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

      █▄                                    ▆                   
  ▃▅▆▅██▆▄▃▃▃▂▂▂▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▆█▅▄▃▃▃▃▃▂▃▂▂▂▂▂▂▂▂ ▃
  7.3 ns          Histogram: frequency by time          186 ns <

julia> import SpecialFunctions

# SpecialFunctions.jl

julia> @benchmark SpecialFunctions.airyai(x) setup=(x=rand()*15.0)
 Range (min … max):   20.394 ns …   1.061 μs  ┊ GC (min … max): 0.00% … 83.60%
 Time  (median):     231.863 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   246.306 ns ± 102.016 ns  ┊ GC (mean ± σ):  0.09% ±  1.35%

                   █   ▃█▄▄▁                                     
  ▄▆▅▂▂▁▁▁▁▁▁▁▁▁▁▁██▇▆▅█████▇▇▆▄▅▄▄▄▃▃▃▃▃▃▃▃▃▃▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂ ▃
  20.4 ns          Histogram: frequency by time          572 ns <

```
The range of times here are important. I'll rearrange to use SIMD in the other PR.

PR #79, opened 2023-03-04
Use explicit SIMD instructions for parallel evaluation of polynomials

heltonmc:
In some ways this is an explicit attempt to solve https://github.com/JuliaMath/Bessels.jl/issues/44 and a more rigorous follow up to https://github.com/JuliaMath/Bessels.jl/pull/45.

We are entirely relying on LLVM emitting vector instructions in the whole library. By my testing this seems to work ok if you have just two polynomials (of low degree) and then you have a single output (functions that return two arguments seem to lose auto-vectorization).

There's probably three types of classes of SIMD polynomial evaluations I can think of. (1) SIMD a single polynomial using either an Estrin or higher order Horner scheme (2) evaluating many polynomials with different coefficients with similar `x` and (3) evaluating the same polynomial (same coefficients) over many values of `x`.

Loop type models (LoopVectorization.jl and friends) do well for the third type of problem. The challenge with special functions is for different values of `x` you have different routines so those type of optimizations are usually challenging without having a dedicated slower SIMD routine compared to the scalar routines we are usually focused on. 

That leaves classes of polynomials for special functions following the first two types. The challenge with the first type is that higher order schemes and Estrin schemes do not give the same floating point numbers due to non-associativity of the floating arithmetic. Generally a straight Horner is more accurate followed by second degree horner. However, this isn't always the case as a second degree horner could actually be more accurate than regular Horner if the coefficients are alternating. This should be tested before use of course but it does have the opportunity to gain performance over regular horner for a high degree polynomial. 

The second class is what this PR focuses on. Essentially, we want to be able to evaluate consecutive `evalpoly(x, Pn)` for the same value `x` with different coefficients `Pn`. In this case we can vectorize separate polynomial evaluations using a straight Horner scheme where we retain the same level of accuracy. Therefore, I don't think there is any compromise in this approach as they numerical values should be equivalent.

Here are some examples of using the module and the performance improvments. 
```julia

using Main.HornerSIMD
using Main.HornerSIMD: LVec

evalpoly_packed(1.2, (LVec{2, Float64}((1.1, 1.2)), LVec{2, Float64}((1.1, 1.2))))
evalpoly_packed(1.2, (LVec{4, Float64}((1.1, 1.2, 1.3, 1.5)), LVec{4, Float64}((1.1, 1.2, 1.1, 1.0))))
evalpoly_packed(1.2, (LVec{8, Float64}((1.1, 1.2, 1.3, 1.5, 1.1, 1.2, 1.3, 1.5)), LVec{8, Float64}((1.1, 1.2, 1.1, 1.0, 1.1, 1.2, 1.1, 1.0))))
evalpoly_packed(VecElement{Float64}(1.2), (LVec{2, Float64}((1.1, 1.2)), LVec{2, Float64}((1.1, 1.2))))



using BenchmarkTools

@inline pack_poly(p1::NTuple{N, T}, p2::NTuple{N, T}, p3::NTuple{N, T}, p4::NTuple{N, T}) where {N, T <: Float64} = ntuple(i -> LVec{4, T}((p1[i], p2[i], p3[i], p4[i])), Val(N))

const NT = 24
const P = (
    ntuple(n -> rand()*(-1)^n / n, NT),
    ntuple(n -> rand()*(-1)^n / n, NT),
    ntuple(n -> rand()*(-1)^n / n, NT),
    ntuple(n -> rand()*(-1)^n / n, NT)
)
const packed_P = pack_poly(P...)

function test(x)
    x2 = x * x
    x3 = x2 * x

    p1 = evalpoly(x3, P[1])
    p2 = evalpoly(x3, P[2])
    p3 = evalpoly(x3, P[3])
    p4 = evalpoly(x3, P[4])

    return muladd(x, -p2, p1), muladd(x2, p4, -p3)
end    

@inline function test_simd(x)
    x2 = x * x
    x3 = x2 * x
    p = @inline evalpoly_packed(x3, packed_P)
    return muladd(x, -p[2].value, p[1].value), muladd(x2, p[4].value, -p[3].value)
end

@inline function test_simd2(x)
    x2 = x * x
    x3 = x2 * x
    p = @inline evalpoly_packed(x3, pack_poly(P...))
    return muladd(x, -p[2].value, p[1].value), muladd(x2, p[4].value, -p[3].value)
end

@inline function test_simd3(x)
    x2 = x * x
    x3 = x2 * x
    p1 = (-0.5253843215045461, 0.15327467952066126, -0.009613425786620758, 0.03472388566919829, -0.006170149922090551, 0.1491658307729879, -0.006127497881439704, 0.09478579099748007, -0.02300284664682352, 0.03765228630052053, -0.004564376925045763, 0.04030818160760519, -0.07423483930698441, 0.005063286843641934, -0.028511541495917378, 0.001286639489236567, -0.04219269322797644, 0.019264463041094002, -0.024846758379011003, 0.04691071886746108, -0.03505574439901283, 0.0036530747019650656, -0.003217711714585987, 0.02980485780160719)
    p2 = (-0.5617359409644243, 0.21639708415293385, -0.1595616502195741, 0.23558871900635361, -0.01711229996212311, 0.1390325911702592, -0.02204364852309128, 0.03762714019980608, -0.02015688493891717, 0.04691788314401733, -0.05284288235650825, 0.059473399369856746, -0.05558293052003346, 0.06908463578987216, -0.028065429488872185, 0.05311837154039087, -0.046791950471468835, 0.05467063178456665, -0.0012121658411338575, 0.04572757730064302, -0.02111992140998074, 0.03368324759867918, -0.02573305661813354, 0.016586970093781076)
    p3 = (-0.5725590283079678, 0.357354857178728, -0.15852680778567496, 0.13799886873237646, -0.13287632961538073, 0.15645986033380913, -0.12663156857906163, 0.056970637917548736, -0.10365230117107738, 0.09851851159110772, -0.07131661552931909, 0.025855405556433803, -0.05470229425565906, 0.003313351497421592, -0.00832867623977364, 0.02660430403564875, -0.05800975963933192, 0.031600276140392486, -0.002263809343951925, 0.0006947072699261092, -0.0283382906800289, 0.01962982236742449, -0.03752538733132978, 0.03500585661769363)
    p4 = (-0.38523819446439866, 0.3816388038108996, -0.07602128517848546, 0.18032013137067715, -0.025836416466692037, 0.04360821045450199, -0.13093109867663433, 0.06350602366713742, -0.09448183067624193, 0.057900124609400694, -0.015088644384017498, 0.07704971041596012, -0.053614493383646246, 0.010833485944489725, -0.001042917623436971, 0.02967905577087363, -0.03263137907617557, 0.03626585702651997, -0.004715722184918155, 0.034381370693092037, -0.01802701761654408, 0.0198574776937958, -0.012362288041529458, 0.005742367480605785)


    p = @inline evalpoly_packed(x3, pack_poly(p1, p2, p3, p4))
    return muladd(x, -p[2].value, p[1].value), muladd(x2, p[4].value, -p[3].value)
end
```

Now the three test functions are different examples  of how this function might be used. In some case the packing of the polynomials can be done at compile time as long as the polynomial degree is known (fixed tuple length). It's also true that some of the values could depend on input parameters but this will slow the routine down if so.

```julia
julia> @benchmark test(x) setup=(x=rand()*2)
 Range (min … max):  29.556 ns … 56.956 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     29.569 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   29.633 ns ±  0.841 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆█▇▃                                                     ▂▂ ▁
  ████▆▁▄▇▆▁▁▄▄▆▅▄▁▁▄▄▅▃▁▃▃▄▄▁▁▄▃▃▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▃▁▁▁▃▁▃▄██ █
  29.6 ns      Histogram: log(frequency) by time        30 ns <

julia> @benchmark test_simd(x) setup=(x=rand()*2)
 Range (min … max):  8.755 ns … 22.459 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.918 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.879 ns ±  0.320 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

             ▂▂                                       ▂█▃▁    
  ▂▂▂▂▂▂▂▃▄▇████▇▄▃▂▂▂▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▂▂▃▄▆████▅▄ ▃
  8.75 ns        Histogram: frequency by time        8.94 ns <

julia> @benchmark test_simd2(x) setup=(x=rand()*2)
 Range (min … max):  8.772 ns … 20.234 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.803 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.811 ns ±  0.205 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                             ▂▃ ▄▆▆▇ ██▆█▇ ▆▅▄▄▂ ▁            
  ▂▂▂▂▂▁▂▂▂▂▂▁▂▂▂▃▃▁▃▄▄▅▅▁▇▇███▁████▁█████▁█████▁██▇▆▅▁▄▄▃▃▃ ▄
  8.77 ns        Histogram: frequency by time        8.82 ns <

julia> @benchmark test_simd3(x) setup=(x=rand()*2)
 Range (min … max):  8.726 ns … 27.596 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.805 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.928 ns ±  0.978 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▃                                                         ▁
  ██▁▁▃▃▁▁▁▁▃▄▁▃▃▁▃▅▅▃▃▄▃▁▃▄▅▄▁▄▃▄▄▃▄▃▃▁▁▁▄▁▄▁▄▁▁▁▁▃▃▁▃▄▃▄▅▆ █
  8.73 ns      Histogram: log(frequency) by time     15.6 ns <

```

First, all three methods are similar and are about 3.5x faster than the non SIMD version. I've tested this on Mac M1 and Intel 8700k. I don't have a computer to test a version for AVX-512 instruction that could pack eight different polynomials. 

There are a couple things I still need to do with the type template and clean up. I don't want to merge this until the Airy functions are ready because this worked stemmed from there. The example test functions are actually the same structure from the Airy functions. I've been trying to reformat those implementations to take advantage of this better but I'm going to get the serial version done there first then come back to editing it for SIMD.


oscardssmith:
I feel like HornerSimd should possibly be a separate package. It seems like it is more broadly useful.

heltonmc:
Yes I was starting to feel the same or perhaps something similar to https://github.com/EnzymeAD/Enzyme.jl/tree/main/lib/EnzymeCore where it is registered but within the package. I've really been hoping to keep everything under the same umbrella here as it's easier for me to keep track of (docs/ci/caching/load times) as I transition to a new job in a couple months.

In any case I do want to take out the second order Horner schemes we already have in some of the files and put them here. I'll probably add a fourth and eighth order scheme as well now that we can depend on these vector instructions being emitted. But ya if there is broad interest after I clean this up then we can move it out / register it separately here.

heltonmc:
But yes... SpecialFunctions.jl is littered with examples where this should be used.. 

https://github.com/JuliaMath/SpecialFunctions.jl/blob/07a890ca813913e7bcc79022a0ca9c64afcd6382/src/gamma_inc.jl#L599-L607

https://github.com/JuliaMath/SpecialFunctions.jl/blob/07a890ca813913e7bcc79022a0ca9c64afcd6382/src/gamma_inc.jl#L791-L794

https://github.com/JuliaMath/SpecialFunctions.jl/blob/07a890ca813913e7bcc79022a0ca9c64afcd6382/src/sincosint.jl#L25-L58

https://github.com/JuliaMath/SpecialFunctions.jl/blob/07a890ca813913e7bcc79022a0ca9c64afcd6382/src/logabsgamma/e_lgamma_r.jl#L139-L141

https://github.com/JuliaMath/SpecialFunctions.jl/blob/07a890ca813913e7bcc79022a0ca9c64afcd6382/src/erf.jl#L409-L428

And that's just the first couple of files I looked at.

In any case this has pretty significant opportunities to speed of complex numbers as well and definitely for a Double64 routine that we could pack.....

src/SIMDMath/arithmetic.jl
```diff
@@ -0,0 +1,44 @@
...
+@inline muladd(x::ScalarTypes, y::LVec{N, T}, z::LVec{N, T}) where {N, T <: FloatTypes} = muladd(constantvector(x, LVec{N, T}), y, z)
+@inline muladd(x::LVec{N, T}, y::ScalarTypes, z::LVec{N, T}) where {N, T <: FloatTypes} = muladd(x, constantvector(y, LVec{N, T}), z)
+@inline muladd(x::ScalarTypes, y::ScalarTypes, z::LVec{N, T}) where {N, T <: FloatTypes} = muladd(constantvector(x, LVec{N, T}), constantvector(y, LVec{N, T}), z)
+@inline muladd(x::LVec{N, T}, y::LVec{N, T}, z::ScalarTypes) where {N, T <: FloatTypes} = muladd(x, y, constantvector(z, LVec{N, T}))
+@inline muladd(x::ScalarTypes, y::LVec{N, T}, z::ScalarTypes) where {N, T <: FloatTypes} = muladd(constantvector(x, LVec{N, T}), y, constantvector(z, LVec{N, T}))
+@inline muladd(x::LVec{N, T}, y::ScalarTypes, z::ScalarTypes) where {N, T <: FloatTypes} = muladd(x, constantvector(y, LVec{N, T}), constantvector(z, LVec{N, T}))
+
+# muladd llvm instructions
+
+@inline function _muladd(x::LVec{2, Float64}, y::LVec{2, Float64}, z::LVec{2, Float64})
```
oscardssmith, line 13: can't you make a length generic version of this?

heltonmc: Should be possible with a generated function though my quick attempt didn't work. Need to improve my meta-programming ability. Most of this is built up by pieces so there is some redundancy that I think I can simplify.

heltonmc: Fully generic version...

```julia
const LLVMType = Dict{DataType, String}(
    Float32  => "float",
    Float64  => "double",
)

 @generated function _muladd(x::LVec{N, T}, y::LVec{N, T}, z::LVec{N, T}) where {N, T <: FloatTypes}
    s = """
        %4 = fmul contract <$N x $(LLVMType[T])> %0, %1
        %5 = fadd contract <$N x $(LLVMType[T])> %4, %2
        ret <$N x $(LLVMType[T])> %5
        """
    return :(
        llvmcall($s, LVec{N, T}, Tuple{LVec{N, T}, LVec{N, T}, LVec{N, T}}, x, y, z)
        )
end
```
```julia
julia> @code_llvm muladd(LVec{5, Float32}((1.1, 1.2, 1.2, 1.3, 1.65)), LVec{5, Float32}((1.1, 1.2, 1.5, 1.3, 1.5)), LVec{5, Float32}((1.1, 1.2, 1.1, 1.2, 1.5)))
;  @ /home/heltonmc/Documents/repos/Bessels_fork.jl/src/SIMDMath/arithmetic.jl:3 within `muladd`
define <5 x float> @julia_muladd_2505(<5 x float> %0, <5 x float> %1, <5 x float> %2) #0 {
top:
; ┌ @ /home/heltonmc/Documents/repos/Bessels_fork.jl/src/SIMDMath/arithmetic.jl:52 within `_muladd`
; │┌ @ /home/heltonmc/Documents/repos/Bessels_fork.jl/src/SIMDMath/arithmetic.jl within `macro expansion`
    %3 = fmul contract <5 x float> %0, %1
    %4 = fadd contract <5 x float> %3, %2
; └└
  ret <5 x float> %4

```

oscardssmith: add Float16 support?

heltonmc: Seems reasonable to me for this. Done.
```julia
julia> @code_llvm muladd(LVec{5, Float16}((1.1, 1.2, 1.2, 1.3, 1.65)), LVec{5, Float16}((1.1, 1.2, 1.5, 1.3, 1.5)), LVec{5, Float16}((1.1, 1.2, 1.1, 1.2, 1.5)))
;  @ /home/heltonmc/Documents/repos/Bessels_fork.jl/src/SIMDMath/arithmetic.jl:3 within `muladd`
define <5 x half> @julia_muladd_435(<5 x half> %0, <5 x half> %1, <5 x half> %2) #0 {
top:
; ┌ @ none within `_muladd`
; │┌ @ none within `macro expansion`
    %3 = fmul contract <5 x half> %0, %1
    %4 = fadd contract <5 x half> %3, %2
; └└
  ret <5 x half> %4
```

src/SIMDMath/readme.md
```diff
@@ -59,22 +59,52 @@ julia> @btime test_simd(x) setup=(x=rand()*2)
...
 The last fact is important because we are actually increasing the total amount of arithmetic operations needed but increasing by a large amount the number of operations that can happen in parallel. The increased operations make the advantages of this approach less straightforward than the first case which is always superior. The second and perhaps most important point is that floating point arithmetic is not associative so these approaches will give slightly different results as we are adding and multiplying in slightly differnet order.
 
+The above benchmarks are as follows.
+```julia
+Julia Version 1.8.2
+Platform Info:
+  OS: macOS (arm64-apple-darwin21.3.0)
+  CPU: 8 × Apple M1
+
+horner_times  = [2.125, 2.125, 2.416, 2.416, 4.000, 12.345, 40.573, 126.353]
```
oscardssmith, line 105: can you normalize these timings?

heltonmc: Yes - good point. I also added a different CPU. Very interesting as it will actually do better than the expected 2x or 4x at high degrees (I'm assuming due to the better memory layout).

For small degrees it is just noisy though so hard to really compare. I'll need a better benchmark I think for this as we are trying to distinguish like one or two multiplies. Though it is cool to see even for really small degrees this matches the regular `evalpoly`.

heltonmc:
Closing in favor of moving to a separate package at https://github.com/heltonmc/SIMDMath.jl. Once that is registered we can depend on that package instead.

PR #84, opened 2023-03-17
Refactor code into separate modules

heltonmc:
As the project grows it might be best to separate namespaces into different modules. The motivation is that the constant polynomials take up a lot of namespace and it might be best to have unique functions into different modules.

In the future if someone wanted to split the package into smaller sub-packages this will help make that transition easier as well as be helpful as I add other functions (Scorer, Struve).

As of now I'm most interested how this affects package load / compile times so I will need to test that and clean some things up.

I think the pathway right now is:
 1. get #79 merged
 2. merge this PR,
 3.  use the SIMD methods for the real Airy functions
 4. Reorganize the complex airy functions for better SIMD (also not depend on Bessel function computation which is slow)
 5. Use the better airy functions in uniform airy type expansions for the Bessel function code
 6. Use those to have routines for complex support in bessel functions

Sounds like a lot but mostly there......


oscardssmith:
I really feel like the SIMD polynomials should probably be a separate package since
1. Other people might want to use them (but not care about bessel functions)
2. Separation of concerns: If someone makes the polynomial code faster, the bessel code doesn't have to change.

heltonmc:
Sounds good to me! I've moved the SIMD code over to https://github.com/heltonmc/SIMDMath.jl and will register later and come back to this in a couple days.

Are we still ok with keeping the module structure here just minus the SIMDMath code?

heltonmc:
Registration at https://github.com/JuliaRegistries/General/pull/79804. I'll come back Monday and update the airy functions with that dependency.

I'm starting to wonder now if we just want to separate out all the functions into different modules (AiryFunctions, GammaFunctions, BesselFunctions, ..etc) and then just reexport them here. I know that was briefly discussed in SpecialFunctions.jl but I would definitely like to have some top line package that contains just julia code that brings together all of the special function libraries. 

Not for sure if all these different packages would be under JuliaMath or a different org (I don't want to spam too much the CI there).

Right now this reorg exports `gamma` but open to ideas. This is a similar discussion to https://github.com/JuliaMath/Bessels.jl/issues/27 where I have gone back and forth....

heltonmc:
Closed in favor of #95

PR #85, opened 2023-03-29
Format besseli for AD and incorporate SIMDMath

heltonmc:
This PR is the first step to generate routines that are AD compatible. The easiest function is `besseli`. @cgeoga this branch should be a good testing bed for AD but as of right now the only function that should be correct is `besseli_positive_args(v, x)` that should be differentiable for any positive argument.

This PR also adds SIMDMath.jl and reformats the U-polynomial completely. Before we formatted the code to rely on the autovectorizer which could do an ok job. Here is a benchmark on the latest tagged branch.

```julia
julia> @benchmark Bessels.Uk_poly20(x, z, y) setup=(x=rand(); z=rand(); y=rand())
 Range (min … max):  32.926 ns … 43.809 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     33.021 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   33.060 ns ±  0.261 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                         ▁▇█▇▇                                 
  ▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▃▄███████▅▃▃▃▃▄▃▃▃▄▄▄▅▄▅▅▄▅▄▄▅▄▄▃▃▄▄▃▃▂ ▃
  32.9 ns         Histogram: frequency by time        33.1 ns <

```

And here is that same benchmark on this branch.
```julia
julia> @benchmark Bessels.Uk_poly20(x, z, y) setup=(x=rand(); z=rand(); y=rand())
 Range (min … max):  18.390 ns … 36.280 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     18.425 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   18.570 ns ±  1.300 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                           ▁
  ███▄▁▁▁▁▁▁▁▁▁▁▁▃▆▅▄▄▃▄▄▁▃▁▃▃▁▁▄▁▃▁▁▁▁▁▃▄▄▃▃▁▁▁▁▄▃▁▁▁▃▁▁▁▁▃▃ █
  18.4 ns      Histogram: log(frequency) by time      25.3 ns <

```
I've optimized the U-polys for AVX instructions where everything is completely unrolled. The llvm generated code is just a string of....
```julia
%12 = fmul contract <4 x double> %5, <double 0.000000e+00, double 0.000000e+00, double 0.000000e+00, double 0xC3BF1EBE423D8BC6>
     %13 = fadd contract <4 x double> %12, <double 0.000000e+00, double 0.000000e+00, double 0x43726F1DD54E5654, double 0x43F62C4DF598A6CA>
     %14 = fmul contract <4 x double> %5, %13
     %15 = fadd contract <4 x double> %14, <double 0.000000e+00, double 0xC3271F57863FBE5A, double 0xC3A8E2CEACC35AF1, double 0xC41D4CC390208D15>
     %16 = fmul contract <4 x double> %5, %15
     %17 = fadd contract <4 x double> %16, <double 0x42DED06D883C7D6C, double 0x435D7B29319145E6, double 0x43CF04AC0BDB47FF, double 0x4437D0394B450FCE>
     %18 = fmul contract <4 x double> %5, %17
     %19 = fadd contract <4 x double> %18, <double 0xC3127D0E84F1180E, double 0xC381446C7CCDC56F, double 0xC3E7A91E93284AA2, double 0xC44AA112D6E4D78F>
     %20 = fmul contract <4 x double> %5, %19
     %21 = fadd contract <4 x double> %20, <double 0x43344560EEC11E37, double 0x43989E3930B3B6E8, double 0x43F8B08BFBA4D477, double 0x4455B4FA5DCA4BB1>
     %22 = fmul contract <4 x double> %5, %21
     %23 = fadd contract <4 x double> %22, <double 0xC34AE03F99BB5229, double 0xC3A7D964D45C558C, double 0xC402A7AFA9C2E24A, double 0xC45AAAEA9F0D29A0>
```
Which is optimal. I then did a fancier version of the split Horner scheme considering a 4th order because the coefficients are already packed into `<4 x double>` vectors. This creates a very good speedup compared to what the autovectorizer was able to do.

In any case - testing `besseli` here for AD should be indicative of all the other functions because it will now use the U asymptotic expansions with SIMDMath, power series, and asymptotic expansion.

Also, with v1.9 dropping soon I think we should update to at least v1.8. `SIMDMath.jl` requires v1.8 though if we want to support v1.6 I can change the compat for SIMDMath


heltonmc:
Ha well got error already using this method with Enzyme hahaha :sweat: 

```julia
function dbesseli_dx(v, x)
    autodiff(Forward, _x->Bessels.besseli_positive_args(v, _x), Duplicated, Duplicated(x, 1.0))[2]
  end
  
  function dbesseli_dv(v, x)
    autodiff(Forward, _v->Bessels.besseli_positive_args(_v, x), Duplicated, Duplicated(v, 1.0))[2]
  end
```
Try to find MWE

heltonmc:
Bisected the error in the large order expansion.... I've submitted the issue over at https://github.com/EnzymeAD/Enzyme.jl/issues/688. I can probably also allow dual numbers on this branch if we want to test ForwardDiff

heltonmc:
Bug in Enzyme should be fixed now (but haven't confirmed).

I've also just released a new version of SIMDMath.jl that can handle SIMD complex polynomial evaluation (complex args / real coefficients). This is needed for the complex evaluation of the uniform asymptotic expansions here. I will update that function so it is able to handle complex numbers.

The last bit to handle for complex numbers is having the uniform airy asymptotic expansions on the imaginary axis at https://dlmf.nist.gov/10.20. @cgeoga Did you have ever attempt those ? I'll work on them but I want to improve the complex airy functions first as they are too slow and optimize them for both `airyai` and `airyaiprime`.

The last bit for complex numbers is in that intermediate range. There's a couple approaches: 1. Backward recurrence with uniform asymptotic expansions, 2. Miller's method, 3. Wronskian with `besselk`.

All three have there drawbacks I believe. The first method will be dictated by the accuracy of the uniform expansions which in general are less accurate. Miller's method is accurate but the amount of terms needed depends on `v` and `x` which doesn't allow very efficient compile time unrolling of loops. It also requires three gamma calls. That makes it worthless for real arguments because power series is numerically stable. But for complex arguments it is much better because gamma calls are for real arguments and it avoids any complex divides (well just one at the end to normalize) so it may be faster for complex arguments.

The wronskian has an issue of needing two besselks and the ratio of besseli (which can be done with an abbreviated form of Miller's method (no gamma calls)) so it is kinda fast but still need besselk which we will probably use the sequence transformations as they seem pretty fast even for complex numbers. But needing two is a downside.

Anyway - I'll try to get each individual algorithm here working for complex arguments, improve the airy functions for complex, then code up the uniform airy expansions. That will give us every algorithm needed for complex domain for all bessel functions. Though will require some effort :)

PR #86, opened 2023-04-04
Improve complex airy asymptotic expansions

heltonmc:
The previous large argument expansions were kind of a mess. They were also much slower than they needed be. By unrolling the expansions to a certain degree we can avoid any complex divisions.

```julia
# latest tagged version
julia> @benchmark Bessels.airyai_large_argument(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 142 evaluations.
 Range (min … max):  705.690 ns …   2.018 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     834.965 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   967.171 ns ± 343.383 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █ ▅ ▃ ▁                                                        
  █▂█▂█▂█▂▇▂▆▂▆▂▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅▂▂▁▂▅▁▂▂▄▂▂▂▄▂▂▁▃▂▁▁▃▂▂▁▃ ▃
  706 ns           Histogram: frequency by time         1.92 μs <

# this PR
julia> @benchmark Bessels.airyaix_large_args(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 952 evaluations.
 Range (min … max):   93.103 ns … 217.356 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      98.878 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   107.704 ns ±  17.242 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆ ▆█▄▂       ▃▆▁ ▆▅                   ▁▁▁                     ▂
  █▇████▇▄▃▄▄▁▃███████▄▄▃▁▃▃▁▁▃▃▄▆▆▇█▆▇▇████▆▅▃▁▄▁▄▃▄▄▄▁▁▁▅▅▇▇▇ █
  93.1 ns       Histogram: log(frequency) by time        188 ns <

```

So in this range the mean time to compute this benchmark is 9x faster. I've also set this up to be able to use SIMDMath to further reduce the time in a different PR. O and these compute both the derivative and the function value where old ones we would need two of those evaluations.

### DIFF
--- a/src/Airy/cairy.jl
+++ b/src/Airy/cairy.jl
@@ -394,6 +394,8 @@ end
 airy_large_argument_cutoff(z::ComplexOrReal{Float64}) = abs(z) > 8.3
 airy_large_argument_cutoff(z::ComplexOrReal{Float32}) = abs(z) > 4
 
+# valid in 0 <= angle(z) <= pi
+# use conjugation for bottom half plane
 function airyai_large_args(z::Complex{T}) where T
     if imag(z) < zero(T)
         out = conj.(airyaix_large_args(conj(z)))
@@ -429,8 +431,6 @@ end
     return ai * Base.FastMath.inv_fast(xsqr) * inv(PIPOW3O2(T)), aip * xsqr * inv(PIPOW3O2(T))
 end
 
-# valid in 0 <= angle(z) <= pi
-# use conjugation for bottom half plane
 @inline function airybix_large_args(z::Complex{T}) where T
     xsqr = sqrt(z)
     xsqrx = Base.FastMath.inv_fast(z * xsqr)
@@ -453,14 +453,13 @@ end
     invx3 = @fastmath inv(z^3)
     p = SIMDMath.horner_simd(invx3, pack_AIRY_ASYM_COEF)
 
-    zvec = SIMDMath.ComplexVec{2, Float64}((xsqrx.re, xsqrx.re), (xsqrx.im, xsqrx.im))
-    zvecn = SIMDMath.ComplexVec{2, Float64}((-xsqrx.re, -xsqrx.re), (-xsqrx.im, -xsqrx.im))
-
     pvec1 = SIMDMath.ComplexVec{2, Float64}((p.re[1], p.re[3]), (p.im[1], p.im[3]))
     pvec2 = SIMDMath.ComplexVec{2, Float64}((p.re[2], p.re[4]), (p.im[2], p.im[4]))
 
-    a = SIMDMath.fmadd(zvec, pvec2, pvec1)
-    b = SIMDMath.fmadd(zvecn, pvec2, pvec1)
+    zvec = SIMDMath.ComplexVec{2, Float64}((xsqrx.re, xsqrx.re), (xsqrx.im, xsqrx.im))
+    zvec = SIMDMath.fmul(zvec, pvec2)
+    a = SIMDMath.fadd(pvec1, zvec)
+    b = SIMDMath.fsub(pvec1, zvec)
 
     A = complex(b.re[1].value, b.im[1].value)
     B = complex(a.re[1].value, a.im[1].value)
--- a/src/Airy/cairy.jl
+++ b/src/Airy/cairy.jl
@@ -451,14 +451,21 @@ end
 
 @inline function compute_airy_asy_coef(z, xsqrx)
     invx3 = @fastmath inv(z^3)
-    p1 = evalpoly(invx3, AIRY_ASYM_COEF[1])
-    p2 = evalpoly(invx3, AIRY_ASYM_COEF[2])
-    p3 = evalpoly(invx3, AIRY_ASYM_COEF[3])
-    p4 = evalpoly(invx3, AIRY_ASYM_COEF[4])
-    A = muladd(xsqrx, -p2, p1)
-    B = muladd(xsqrx, p2, p1)
-    C = muladd(xsqrx, -p4, p3)
-    D = muladd(xsqrx, p4, p3)
+    p = SIMDMath.horner_simd(invx3, pack_AIRY_ASYM_COEF)
+
+    zvec = SIMDMath.ComplexVec{2, Float64}((xsqrx.re, xsqrx.re), (xsqrx.im, xsqrx.im))
+    zvecn = SIMDMath.ComplexVec{2, Float64}((-xsqrx.re, -xsqrx.re), (-xsqrx.im, -xsqrx.im))
+
+    pvec1 = SIMDMath.ComplexVec{2, Float64}((p.re[1], p.re[3]), (p.im[1], p.im[3]))
+    pvec2 = SIMDMath.ComplexVec{2, Float64}((p.re[2], p.re[4]), (p.im[2], p.im[4]))
+
+    a = SIMDMath.fmadd(zvec, pvec2, pvec1)
+    b = SIMDMath.fmadd(zvecn, pvec2, pvec1)
+
+    A = complex(b.re[1].value, b.im[1].value)
+    B = complex(a.re[1].value, a.im[1].value)
+    C = complex(b.re[2].value, b.im[2].value)
+    D = complex(a.re[2].value, a.im[2].value)
     return A, B, C, D
 end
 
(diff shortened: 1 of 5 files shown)


heltonmc:
I added two instances of fastmath only on the single divisions to compute inverse. This doesn't affect the accuracy but has substantial effect on timing.

```julia
#no fast
julia> @benchmark Bessels.airybix_large_args(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 932 evaluations.
 Range (min … max):  107.608 ns … 208.797 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     119.296 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   117.900 ns ±   4.166 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                          ▇                 █  ▁  ▂▆▇▆▂  ▂▁      
  ▂▂▂▁▁▁▁▁▁▂▃▅▃▂▁▁▂▄▇█▃▂▁▃██▃▂▂▂▂▁▂▂▁▂▂▂▁▁▁▃█▆██▃▅█████▅███▅▂▂▂ ▄
  108 ns           Histogram: frequency by time          122 ns <

# fast_inv
julia> @benchmark Bessels.airybix_large_args(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 956 evaluations.
 Range (min … max):   90.730 ns … 204.802 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     101.634 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   101.366 ns ±  10.150 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▁   ▄█▇                                                     
  ▂▂▄█▇▂▂███▄▂▂▂▂▂▂▂▂▂▂▁▂▁▂▁▁▁▁▁▂▂▁▁▁▁▂▁▁▁▂▁▁▂▁▂▁▂▁▂▁▁▁▂▁▁▁▂▁▁▂ ▃
  90.7 ns          Histogram: frequency by time          170 ns <

```
So about ~17-20 nanoseconds here.

src/Airy/cairy.jl
```diff
@@ -377,197 +377,98 @@ end
...
-    if x < zero(T) && y < 5
-        out = d + im*c
-        check_conj && (out = conj(out))
-        return out
+@inline function airyaix_large_args(z::Complex{T}) where T
+    xsqr = sqrt(z)
+    xsqrx =  Base.FastMath.inv_fast(z * xsqr)
+    A, B, C, D = compute_airy_asy_coef(z, xsqrx)
+    
+    if abs(angle(z)) > 4π/6
```
oscardssmith, line 419: rewrite in terms of ratio of components. `angle` is slow. (i.e. something like `if abs(real(z)) < sqrt(3)*abs(imag(z))`)

src/Airy/cairy.jl
```diff
@@ -377,197 +377,98 @@ end
...
-        abs(t) < tol*abs(out) && break
-        t *= 3*(i + one(T)/6) * (i + T(5)/6) / (a*(i + one(T)))
+# valid in 0 <= angle(z) <= pi
+# use conjugation for bottom half plane
+@inline function airybix_large_args(z::Complex{T}) where T
+    xsqr = sqrt(z)
+    xsqrx = Base.FastMath.inv_fast(z * xsqr)
+    A, B, C, D = compute_airy_asy_coef(z, xsqrx)
+    
+    if abs(angle(z)) < 2π/3
```
oscardssmith, line 439: same here.

src/Airy/cairy.jl
```diff
@@ -377,197 +377,98 @@ end
...
-    a = 4*xsqr*x
-    for i in 0:MaxIter
-        out += t
-        abs(t) < tol*abs(out) && break
-        t *= 3*(i - one(T)/6) * (i + T(7)/6) / (a*(i + one(T)))
-    end
-    return -out * exp(a / 6) * sqrt(xsqr) / sqrt(T(π)^3)
+@inline function compute_airy_asy_coef(z, xsqrx)
+    invx3 = @fastmath inv(z^3)
+    p1 = evalpoly(invx3, AIRY_ASYM_COEF[1])
```
oscardssmith, line 454: SIMDMath here?

heltonmc: Ya might as well make the dependency now !! (I set this up so it will be easy switch). I think the complex SIMD_evalpoly method is working now.

src/Airy/cairy.jl
heltonmc: There might be a better way to do it but it did shave off ~10ns

heltonmc:
I've added SIMDMath here. Need to resolve the version dependencies etc on julia and documentation. I'm fine bumping Bessels require julia v1.8. Part of me thinks we should just require v1.9 soon when that drops as precompilation is so powerful for this package.

Anway, I need to add some methods in SIMDMath - mainly methods for `fnmuladd` for complex numbers so I can switch that over to here.

But if you want to see the true awesomeness of that package....

```julia

# old benchmark
julia> @benchmark Bessels.airyaix_large_args(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 952 evaluations.
 Range (min … max):   93.103 ns … 217.356 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      98.878 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   107.704 ns ±  17.242 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆ ▆█▄▂       ▃▆▁ ▆▅                   ▁▁▁                     ▂
  █▇████▇▄▃▄▄▁▃███████▄▄▃▁▃▃▁▁▃▃▄▆▆▇█▆▇▇████▆▅▃▁▄▁▄▃▄▄▄▁▁▁▅▅▇▇▇ █
  93.1 ns       Histogram: log(frequency) by time        188 ns <

# with SIMDMath
julia> @benchmark Bessels.airyaix_large_args(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 985 evaluations.
 Range (min … max):  54.362 ns … 123.783 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     56.106 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   62.405 ns ±  10.390 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▁  ▆▁                                               ▂▄▇     ▂
  ██▄▄██▆▄▇▆▃▅▃▃▄▅▄▅▃▃▅▄▃▁▃▁▄▄▁▄▁▁▃▄▁▁▃▁▁▁▃▃▁▁▁▃▁▇█▃▁▁▅████▄▅█ █
  54.4 ns       Histogram: log(frequency) by time      78.4 ns <

```

So ya - that's computing the asymptotic expansion in entire complex seconds in like 50 nanoseconds...

If you want to see the SpecialFunctions comparison of same benchmark.....

```julia
julia> import SpecialFunctions

julia> @benchmark SpecialFunctions.airyaix(x) setup=(x=(8.3 + rand()*2)*cispi(2*rand()))
BenchmarkTools.Trial: 10000 samples with 226 evaluations.
 Range (min … max):  328.677 ns …   2.421 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):       1.386 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.059 μs ± 498.986 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █▆                                                  ▃▄▂▁      
  ▄██▄▂▂▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▄▂▁▂▁▁▁▁▁▁▁▁▂▅█████▇▆▄▄ ▃
  329 ns           Histogram: frequency by time         1.54 μs <

```
So 25x faster....

src/Airy/cairy.jl
heltonmc: See comment below. Need to tag another version of SIMDMath to better handle negated muladd for complex numbers

heltonmc:
I think ready to merge now. I've updated all CI and compat to julia v1.8. I think let's just plan for a big v0.3.0 release which by that time julia v1.9 should definitely have dropped.

I'll work next on improving the power series for the complex airy functions and then the intermediate range (probably going to use the sequence transformation from other thread). That should significantly improve performance in other regions.

PR #87, opened 2023-04-06
Unroll power series in complex airy functions

heltonmc:
This continues the improvement to the airy functions by speeding up the power series.

Main improvements are 1. unrolling the loop to specified terms (prevents tolerance checks and expensive `abs` in complex plane) 2. using SIMDMath for vectorizing, 3. avoiding any complex divisions.

Accuracy for these functions should be the same (the simple difference is recursion vs unrolling).

Benchmarks indicate these functions are 11-15x faster.

```julia
# latest tagged version
julia> @benchmark Bessels.airyai_power_series(z) setup=(z=rand()*8*cispi(rand()))
BenchmarkTools.Trial: 10000 samples with 987 evaluations.
 Range (min … max):   42.569 ns … 891.616 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     335.094 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   331.138 ns ± 134.109 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▁ ▂ ▄ ▄▅ ▆ ▆   ▁ ▇ ▅ ▇ ▇ ▅▃ ▇ ▆ ▆ ▆ ▇ █ ▅ ▄▂ ▅ ▄ ▁       
  ▂▁▄▁▆▂█▁█▁█▂██▃█▃█▂▂▂███▂█▂█▂█▄██▅█▃█▃█▂█▂█▂█▃█▄██▅█▄█▂█▂▆▂▆▂ ▅
  42.6 ns          Histogram: frequency by time          584 ns <

julia> @benchmark Bessels.airybi_power_series(z) setup=(z=rand()*8*cispi(rand()))
BenchmarkTools.Trial: 10000 samples with 979 evaluations.
 Range (min … max):   41.523 ns … 605.067 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     334.922 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   326.983 ns ± 132.780 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▂ ▄ ▅ ▇ ▅ ▇ ▇   ▆ █ █ ▇ ▆ ▃▂ ▇ ▅ ▇ ▇ █ ▇ ▇ █ ▆ ▇ ▄ ▁     
  ▂▂▄▁▇▁█▂█▂█▂█▂█▂█▂█▂▁▁█▃█▂█▃█▄█▅██▇█▃█▄█▃█▂█▂█▂█▂█▂█▃█▂█▂█▂▆▃ ▅
  41.5 ns          Histogram: frequency by time          569 ns <

# this PR
julia> @benchmark Bessels.airyai_power_series(z) setup=(z=rand()*8*cispi(rand()))
 Range (min … max):  28.975 ns … 45.823 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     29.270 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   29.321 ns ±  0.599 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

         █▁                                                    
  ▂▂▁▂▂▂▃██▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂ ▂
  29 ns           Histogram: frequency by time        31.2 ns <

julia> @benchmark Bessels.airybi_power_series(z) setup=(z=rand()*8*cispi(rand()))
 Range (min … max):  29.051 ns … 39.968 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     29.204 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   29.232 ns ±  0.364 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                      ▁▁▄▃█▅█▅█▄▄▄             
  ▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▃▃▃▅▅▇██████████████▅▆▄▄▃▃▂▂▂ ▄
  29.1 ns         Histogram: frequency by time        29.3 ns <

```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,7 @@ SIMDMath = "5443be0b-e40a-4f70-a07e-dcd652efc383"
 
 [compat]
 julia = "1.8"
-SIMDMath = "0.2.1"
+SIMDMath = "0.2.3"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/src/Airy/cairy.jl
+++ b/src/Airy/cairy.jl
@@ -259,8 +259,8 @@ function airyai_power_series(z::Complex{T}) where T
     z2 = z * z
     z3 = z2 * z
     p = SIMDMath.horner_simd(z3, pack_AIRYAI_POW_COEF)
-    ai = muladd(-complex(p.re[2].value, p.im[2].value), z, complex(p.re[1].value, p.im[1].value))
-    aip = muladd(complex(p.re[4].value, p.im[4].value), z2, -complex(p.re[3].value, p.im[3].value))
+    ai = muladd(-p[2], z, p[1])
+    aip = muladd(p[4], z2, -p[3])
     return ai, aip
 end
 
@@ -268,8 +268,8 @@ function airybi_power_series(z::Complex{T}) where T
     z2 = z * z
     z3 = z2 * z
     p = SIMDMath.horner_simd(z3, pack_AIRYBI_POW_COEF)
-    bi = muladd(complex(p.re[2].value, p.im[2].value), z, complex(p.re[1].value, p.im[1].value))
-    bip = muladd(complex(p.re[4].value, p.im[4].value), z2, complex(p.re[3].value, p.im[3].value))
+    bi = muladd(p[2], z, p[1])
+    bip = muladd(p[4], z2, p[3])
     return bi, bip
 end
 
@@ -368,10 +368,7 @@ end
     a = SIMDMath.fadd(pvec1, zvec)
     b = SIMDMath.fsub(pvec1, zvec)
 
-    A = complex(b.re[1].value, b.im[1].value)
-    B = complex(a.re[1].value, a.im[1].value)
-    C = complex(b.re[2].value, b.im[2].value)
-    D = complex(a.re[2].value, a.im[2].value)
+    A, B, C, D = b[1], a[1], b[2], a[2]
     return A, B, C, D
 end
 


heltonmc:
Right now I switched everything to be promoted to Float64 at the top function call so everything is internally being done in Float64.

Of course, this isn't entirely needed for Float32 where we only need the asymptotic expansion and the power series to cover all numbers. I think having a dedicated routine for that would be helpful of course but the power series would still need to be done in Float64 precision to avoid severe roundoff. I believe the asymptotic expansion could be done in Float32 but because we are stretching the range down to use x > 4 we would need a similar number of terms.

Either way I'll try to add dedicated routine.

heltonmc:
I put in the Float32 branches so that they will hit the power series or asymptotic expansion.

Here is the spot of worst branch.
```julia
julia> @benchmark Bessels.airyai(z) setup=(z=4.0 + rand() + 2.0*rand()*im)
 Range (min … max):  1.351 μs …  2.577 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.453 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.460 μs ± 45.744 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                ▂▆▅▃  ▅▇█▅▂ ▃▄▅▅▁  ▂▃                         
  ▁▁▁▁▁▁▁▂▃▄▅▇▄▅████▇██████▇█████▆████▆▄▄▄▄▃▃▂▂▃▃▃▂▂▂▁▂▂▁▁▁▁ ▄
  1.35 μs        Histogram: frequency by time        1.59 μs <

julia> @benchmark Bessels.airyai(z) setup=(z=Float32(4.0 + rand()) + Float32(2.0*rand())*im)
BenchmarkTools.Trial: 10000 samples with 957 evaluations.
 Range (min … max):  89.520 ns … 200.393 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     93.322 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   93.651 ns ±   8.953 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂▇▄                              ▆█▄▁ ▇▇▁                    ▂
  ███▄▁▁▁▁▁▁▁▁▁▁▁▁▁▅▇▆▄▄▃▁▃▃▁▁▁▃▁▁▃████████▁▃██▄▁▁▁▃██▆▅▆██▆▄▄ █
  89.5 ns       Histogram: log(frequency) by time      96.1 ns <

```
So Float32 hits the other branch where Float64 has to hit the continued fraction and recurrence which is VERY slow. Next PR will put in a new different algorithm for that branch.

heltonmc:
I've updated to use latest version of SIMDMath which allows indexing into the Vec which is much cleaner for both real and complex numbers. Performance differences was negligible though this was slightly <~1ns faster.

PR #88, opened 2023-04-07
better sin_sum

oscardssmith:
This makes `sin_sum` work for summing an arbitrary number of elements and retain accuracy for large inputs (>`1e15`), and uses this improvement to remove the edge cases in `bessely` and `besselj`. Performance and accuracy information to come.

### DIFF
--- a/src/besselj.jl
+++ b/src/besselj.jl
@@ -63,7 +63,7 @@ function _besselj0(x::Float64)
 
         a = SQ2OPI(T) * sqrt(xinv) * p
         xn = xinv*q
-        b = sin_sum((x,), (pi/4, xn))
+        b = sin_sum(x, pi/4, xn)
         return a * b
     end
 end
@@ -131,7 +131,7 @@ function _besselj1(x::Float64)
 
         a = SQ2OPI(T) * sqrt(xinv) * p
         xn = xinv*q
-        b = sin_sum((x,), (-pi/4, xn))
+        b = sin_sum(x, -pi/4, xn)
         return a * b * s
     end
 end
--- a/src/besselj.jl
+++ b/src/besselj.jl
@@ -59,21 +59,11 @@ function _besselj0(x::Float64)
             q2 = (-1/8, 25/384, -1073/5120, 375733/229376)
             p = evalpoly(x2, p2)
             q = evalpoly(x2, q2)
-            if x > 1e15
-                a = SQ2OPI(T) * sqrt(xinv) * p
-                xn = muladd(xinv, q, -PIO4(T))
-                s_x, c_x = sincos(x)
-                s_xn, c_xn = sincos(xn)
-                return a * (c_x * c_xn - s_x * s_xn)
-            end
         end
 
         a = SQ2OPI(T) * sqrt(xinv) * p
-        xn = muladd(xinv, q, -PIO4(T))
-
-        # the following computes b = cos(x + xn) more accurately
-        # see src/misc.jl
-        b = cos_sum(x, xn)
+        xn = xinv*q
+        b = sin_sum((x,), (pi/4, xn))
         return a * b
     end
 end
@@ -137,21 +127,11 @@ function _besselj1(x::Float64)
             q2 = (3/8, -21/128, 1899/5120, -543483/229376)
             p = evalpoly(x2, p2)
             q = evalpoly(x2, q2)
-            if x > 1e15
-                a = SQ2OPI(T) * sqrt(xinv) * p
-                xn = muladd(xinv, q, -3 * PIO4(T))
-                s_x, c_x = sincos(x)
-                s_xn, c_xn = sincos(xn)
-                return s * a * (c_x * c_xn - s_x * s_xn)
-            end
         end
 
         a = SQ2OPI(T) * sqrt(xinv) * p
-        xn = muladd(xinv, q, -3 * PIO4(T))
-
-        # the following computes b = cos(x + xn) more accurately
-        # see src/misc.jl
-        b = cos_sum(x, xn)
+        xn = xinv*q
+        b = sin_sum((x,), (-pi/4, xn))
         return a * b * s
     end
 end
@@ -182,7 +162,7 @@ end
 #####
 
 besselj0(z::T) where T <: Union{ComplexF32, ComplexF64} = besseli0(z/im)
-besselj1(z::T) where T <: Union{ComplexF32, ComplexF64} = besseli1(z/im) * im 
+besselj1(z::T) where T <: Union{ComplexF32, ComplexF64} = besseli1(z/im) * im
 
 
 #                  Bessel functions of the first kind of order nu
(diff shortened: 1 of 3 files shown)


heltonmc:
Probably also worth switching out 
https://github.com/JuliaMath/Bessels.jl/blob/a94bb7f854985c7f97ee24760cf25b0b23c9754d/src/asymptotics.jl#L40-L52

I haven't tested thoroughly if that section is the cause or not but this is related to https://github.com/JuliaMath/Bessels.jl/issues/37

oscardssmith:
To do that properly I would have to restructure this a little bit to make it possible to get the sin and cos sum simultaneously.

heltonmc:
Ya I think the error is coming before that anyway but it's been a while since I've looked at that function. This is great on my tests the accuracy was similar and was faster.

I do think this function is generally useful enough to be in base because so many people naively will compute sin(x + y) which loses lots of precision.... but I'm obviously bias :smile:

oscardssmith:
I understand where you're coming from, but the problem of errors not being composable isn't really solvable by adding random functions to base.

heltonmc (approved):

heltonmc:
Ps. Totally fine to merge whenever you are ready!

oscardssmith:
I want to do some benchmarks on the 24 to 1e15 region first to make sure I haven't regressed anything.

oscardssmith:
I also think I can get rid of the two separate inputs which would be a bit cleaner.

oscardssmith:
I've done the benchmarking and cleanup that I wanted, so I think this is now good to go. The only thing I'm not 100% happy with in this version is that `sin_sum` of Float32s will now lose a little accuracy since it accumulates into a Float64 which can incur some inaccuracy since `floatmax(Float32)` is `3.4f38` (but doing the reduction this way is ~20x faster than the  method this uses for Float64.

oscardssmith:
OK, I'm finally relatively happy with the Float32 version of the reduction. It still can lose some precision, but it's a lot better than before while maintaining relatively decent speed when you don't have large arguments.

src/besselj.jl
heltonmc, line 1: I noticed that the Float32 implementations still use `p = p * cos(xn + x)`. Should this be switched over to the better sum formula now?

oscardssmith: possibly. it will be a bit slower, so it probably needs a benchmark and accuracy plot to determine.

heltonmc: I can check it out tomorrow ! This weekend was busy with family but have time to look tomorrow more carefully

heltonmc: Yeesh - the besselj0 implementation was some of my first julia code which has been improved but the tests aren't very robust.

```julia
julia> Bessels.besselj0(328049.34f0)
-0.0013240778f0

julia> Bessels.besselj0(Float64(328049.34f0))
-0.0013258623831875669
```
I'm imagining the better sin sum here could help.

heltonmc:
@oscardssmith I was fixing the Float32 case (which drastically improves accuracy) but was noticing that the garbage collector was running. 
For Float64 case.

```julia
# master
julia> @benchmark besselj0(x) setup=(x=rand()*50 + 15.0)
 Range (min … max):  28.635 ns …  1.378 μs  ┊ GC (min … max): 0.00% … 94.85%
 Time  (median):     65.963 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   60.570 ns ± 44.076 ns  ┊ GC (mean ± σ):  2.31% ±  3.27%

  ▇    ▄                          ▂▇█▅▂          ▁▁▁      ▁▁▁ ▂
  █▁▆▄▁█▁▄▃▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█████▇▆▃▃▁▁▄▁▁▇███▇▆▆▄▅████ █
  28.6 ns      Histogram: log(frequency) by time      93.2 ns <

```

Then the Float32 case I was actually getting allocations.
```julia
# on my test branch
julia> @benchmark besselj0(x) setup=(x=Float32(rand()*50 + 15.0))
BenchmarkTools.Trial: 10000 samples with 982 evaluations.
 Range (min … max):  60.338 ns …  1.372 μs  ┊ GC (min … max): 0.00% … 95.17%
 Time  (median):     63.087 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   66.210 ns ± 48.384 ns  ┊ GC (mean ± σ):  3.00% ±  3.91%

  ▁▄▅▇██▇▆▄▂▁                      ▁▁                   ▁▁▁   ▂
  ████████████▇▅▃▄▁▁▁▁▁▃▁▁▁▁▁▁▃▃▁▅█████▇▆▆▄▅▅▅▄▃▁▁▁▅▆▆██████▇ █
  60.3 ns      Histogram: log(frequency) by time      90.4 ns <

 Memory estimate: 80 bytes, allocs estimate: 4.
```

The benchmarks on tagged version
```julia
# Bessels v0.2.8
julia> @benchmark besselj0(x) setup=(x=Float32(rand()*50 + 15.0))
 Range (min … max):  15.738 ns … 28.151 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     16.074 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.054 ns ±  0.319 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                         ▂▄▂                ▁▆█▄               
  ▁▁▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▃▅███▅▃▁▁▁▁▁▁▂▂▂▃▄▅▆▇████▆▃▂▂▃▄▅▆▆▆▄▃▂ ▃
  15.7 ns         Histogram: frequency by time        16.2 ns <

julia> @benchmark besselj0(x) setup=(x=rand()*50 + 15.0)
 Range (min … max):  16.520 ns … 32.577 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     16.916 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.940 ns ±  1.976 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▅ ▂    ▁                                    ▆               
  ██▂█▇▁▁▁█▆▂▁▂▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▂▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▃▂▁▃ ▃
  16.5 ns         Histogram: frequency by time        22.6 ns <

```

oscardssmith:
Presumably you introduced a type instability somewhere.

heltonmc:
Ill look closer but I get the allocation pulling out the `rem_pio2_sum`

```julia

using Base.Math: sin_kernel, cos_kernel, sincos_kernel, rem_pio2_kernel, DoubleFloat64, DoubleFloat32
function rem_pio2_sum(xs::Vararg{Float64})
    n = 0
    hi, lo = 0.0, 0.0
    for x in xs
        if abs(x) <= pi/4
            s = x + hi
            lo += (x - (s - hi))
        else
            n_i, y = rem_pio2_kernel(x)
            n += n_i
            s = y.hi + hi
            lo += (y.hi - (s - hi)) + y.lo
        end
        hi = s
    end
    while hi > pi/4
        hi -= pi/2
        lo -= 6.123233995736766e-17
        n += 1
    end
    while hi < -pi/4
        hi += pi/2
        lo += 6.123233995736766e-17
        n -= 1
    end
    return n, DoubleFloat64(hi, lo)
end
```
```
julia> @benchmark rem_pio2_sum(x, y, z) setup=(x=rand(); y = rand(); z = rand())
 Range (min … max):  26.606 ns …  1.319 μs  ┊ GC (min … max): 0.00% … 95.97%
 Time  (median):     29.220 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   32.298 ns ± 47.517 ns  ┊ GC (mean ± σ):  6.02% ±  4.00%

   ▃▅████▄▃▅▅▃ ▁▁                     ▁                  ▁▁▁▁ ▂
  ███████████████▆▄▁▁▁▃▁▁▁▁▁▁▁▁▁▄▄▆▇▇█████▇▆▇▆▆▄▄▅▄▁▄▁▆▇▆████ █
  26.6 ns      Histogram: log(frequency) by time      54.7 ns <

 Memory estimate: 80 bytes, allocs estimate: 4.
```

PR #94, opened 2023-04-24
Use Levin transform in Complex airy functions

heltonmc:
This PR finally finishes off the complex evaluation of the airy functions. The algorithms come down to
1. Power series for small args
2. Asymptotic expansion for large args
3. Levin transform of asymptotic expansion (when asymptotic expansion is an alternating series)
4. Phase rotation

The final speedup is very large...
```julia
# old
julia> @benchmark Bessels.airyai(z) setup=(z=cispi(rand())*20.0*rand())
BenchmarkTools.Trial: 7616 samples with 985 evaluations.
 Range (min … max):   52.853 ns …   3.171 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     517.425 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   665.591 ns ± 450.075 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

              █ ▃▂                                               
  ▂▃▄▅▅▃▄▅▄▄▄▄█▆██▇▆▃▃▄▃▃▂▃▃▄▃▂▄▂▃▂▂▃▁▃▂▂▃▃▄▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  52.9 ns          Histogram: frequency by time         2.06 μs <

# new
julia> @benchmark Bessels.airyai(z) setup=(z=cispi(rand())*20.0*rand())
 Range (min … max):   33.650 ns … 471.640 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      91.471 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   100.555 ns ±  69.105 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇▁▃    ▅█ ▇             ▂▅▄       ▁                           ▂
  ███▁▁▁▁████▇▃▁▁▁▁▁▁▁▁▁▁▁███▇▄▁▁▁▁▇█▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅█ █
  33.6 ns       Histogram: log(frequency) by time        462 ns <

```
It's hard to quantify exactly how much faster this is compared to the previous because there are so many different ranges but paying attention here to the max range (471 compared to 3171) and the mean and std is relevant.

These implementations are also more accurate! 

Lot's to clean up obviously as this has been going on for a while along with several changes to master that need to be resolved. I'll try to keep working on this over next few days. Want to tag @cgeoga on this as well because it implements the Levin transform (which is in the same form as needed for `besselk`). The algorithm works especially well for phase < pi/2 the issue is that there is a change in the asymptotic forms outside of that so the series is no longer strictly alternating.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -4,7 +4,7 @@ using Base.Math: sin_kernel, cos_kernel, sincos_kernel, rem_pio2_kernel, DoubleF
     computes sin(sum(xs)) where xs are sorted by absolute value
     Doing this is much more accurate than the naive sin(sum(xs))
 """
-function sin_sum(xs::Vararg{T})::T where T<:Base.IEEEFloat
+function sin_sum(xs::Vararg{T, N})::T where {T<:Base.IEEEFloat, N}
     n, y = rem_pio2_sum(xs...)
     n &= 3
     if n == 0
@@ -22,7 +22,7 @@ end
     computes sincos(sum(xs)) where xs are sorted by absolute value
     Doing this is much more accurate than the naive sincos(sum(xs))
 """
-function sincos_sum(xs::Vararg{T})::T where T<:Base.IEEEFloat
+function sincos_sum(xs::Vararg{T, N})::T where {T<:Base.IEEEFloat, N}
     n, y = rem_pio2_sum(xs...)
     n &= 3
     si, co = sincos_kernel(y)
@@ -37,7 +37,7 @@ function sincos_sum(xs::Vararg{T})::T where T<:Base.IEEEFloat
     end
 end
 
-function rem_pio2_sum(xs::Vararg{Float64})
+function rem_pio2_sum(xs::Vararg{Float64, N}) where N
     n = 0
     hi, lo = 0.0, 0.0
     for x in xs
@@ -65,7 +65,7 @@ function rem_pio2_sum(xs::Vararg{Float64})
     return n, DoubleFloat64(hi, lo)
 end
 
-function rem_pio2_sum(xs::Vararg{Float32})
+function rem_pio2_sum(xs::Vararg{Float32, N}) where N
     y = 0.0
     n = 0
     # The minimum cosine or sine of any Float32 that gets reduced is 1.6e-9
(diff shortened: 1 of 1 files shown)


src/Airy/cairy.jl
```diff
@@ -56,29 +56,47 @@ function _airyai(z::Complex{T}) where T <: Union{Float32, Float64}
...
+    check_conj = false
+    if y < zero(T)
+        z = conj(z)
+        check_conj = true
+    end
+
+    if airy_large_argument_cutoff(x, y)
+        r = airyaix_large_args(z)[1] * exp(-2 * z * sqrt(z) / 3)
+    elseif airyai_levin_cutoff(x, y)
+        r = airyaix_levin(z, Val(17)) * exp(-2 * z * sqrt(z) / 3)
```
oscardssmith, line 71: `T(-2/3)` to save a division.

src/Airy/cairy.jl
```diff
@@ -56,29 +56,47 @@ function _airyai(z::Complex{T}) where T <: Union{Float32, Float64}
...
+    elseif airyai_power_series_cutoff(x, y)
+        r = airyai_power_series(z)[1]
+    else
+        r = cispi(one(T)/3) * _airyai(-z*cispi(one(T)/3))  + cispi(-one(T)/3) * _airyai(-z*cispi(-one(T)/3))
```
oscardssmith, line 75: you should check that the compiler is smart enough here (and if not manually tell it that `cispi(1/3) = conj(cispi(-1/3))`

src/math.jl
```diff
@@ -0,0 +1,120 @@
+# function to more accurately compute cos(x + xn)
+# see https://github.com/heltonmc/Bessels.jl/pull/13
+# written by @oscardssmith
+function cos_sum(x, xn)
```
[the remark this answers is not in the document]
heltonmc: Ya needs a rebase. I'll try to figure that out

PR #95, opened 2023-04-24
Use Levin transform in Complex airy functions (again)

heltonmc:
Supersedes #94 which includes the rebase.


src/Airy/cairy.jl
```diff
@@ -50,35 +50,55 @@ airyai(z::Number) = _airyai(float(z))
...
-            xabs = abs(x)
-            xx = 2 * xabs * sqrt(xabs) / 3
-            Jv, Yv = besseljy_positive_args(one(T)/3, xx)
-            Jmv = (Jv - sqrt(T(3)) * Yv) / 2
-            return convert(eltype(z), sqrt(xabs) * (Jmv + Jv) / 3)
-        else
-            return cispi(one(T)/3) * _airyai(-z*cispi(one(T)/3))  + cispi(-one(T)/3) * _airyai(-z*cispi(-one(T)/3))
-        end
+        c = cispi(one(T)/3)
+        r =  exp(2 * z * sqrt(z) / 3) * (c * _airyai(-z*c)  + conj(c) * _airyai(-z*conj(c)))
```
oscardssmith, line 99: `T(2/3)`

src/Airy/cairy.jl
```diff
@@ -111,29 +131,48 @@ function _airyaiprime(z::Complex{T}) where T <: Union{Float32, Float64}
...
+    end
+
+    if airy_large_argument_cutoff(x, y)
+        r = airyaix_large_args(z)[2]
+    elseif airyai_levin_cutoff(x, y)
+        r = airyaiprimex_levin(z, Val(17))
+    elseif airyai_power_series_cutoff(x, y)
+        r = airyai_power_series(z)[2] * exp(T(2)/3 * z * sqrt(z))
+    else
+        r = exp(2 * z * sqrt(z) / 3) * (-cispi(T(2)/3) * _airyaiprime(-z*cispi(one(T)/3))  - cispi(-T(2)/3) * _airyaiprime(-z*cispi(-one(T)/3)))
```
oscardssmith, line 172: `T(2/3)`

src/Airy/cairy.jl
```diff
@@ -372,20 +410,57 @@ end
     return A, B, C, D
 end
 
+@generated function airyaix_levin(x::Complex{T}, ::Val{N}) where {T <: Union{Float32, Float64}, N}
+    :(
+        begin
+            xsqr = sqrt(x)
+            out = zero(typeof(x))
+            t = GAMMA_ONE_SIXTH(T) * GAMMA_FIVE_SIXTHS(T) / 4
+            a = inv(4*xsqr*x)
```
oscardssmith, line 419: `4/(xsqr*x)`

src/Airy/cairy.jl
```diff
@@ -372,20 +410,57 @@ end
...
+
+            l = @ntuple $N i -> begin
+                out += t
+                t *= -3 * a * (i - 5//6) * (i - 1//6) / i
```
oscardssmith, line 423: `t *= a * (i - 5//6) * muladd(1/2, i,  -1)`

src/Airy/cairy.jl
```diff
@@ -372,20 +410,57 @@ end
...
+                invt = @fastmath inv(t)
+                Vec{4, T}((reim(out * invt)..., reim(invt)...))
+            end
+            return @fastmath levin_transform(l) / (sqrt(T(π)^3) * sqrt(xsqr))
```
oscardssmith, line 427: `T(π)^(3//2)`  should also constant fold and might be a bit more accurate.

src/Airy/cairy.jl
heltonmc: I think it would have to be `t *= a * (i - 5//6) * muladd(1/2, 1/i,  -3)` but on a quick test it wasn't as accurate would have to check whole range though to be sure.

oscardssmith: wolfram suggests `a*(3 - 3*i - (5/12)/i)`

src/Airy/cairy.jl
heltonmc: Could be mistaken but are those equivalent ? Would you prefer `a = 1 / (4*xsqr*x)`? Or `0.25 / (xsqr*x)`?

src/Airy/cairy.jl
heltonmc: Seems slightly more accurate will make that change.

heltonmc: Also has advantage of one complex multiply instead of perhaps a complex and two scalar multiples depending if the compiler rearranged that at all

src/Airy/cairy.jl
oscardssmith: oh, duh. `.25/...` (it's very minor but saves a multiplication)

heltonmc:
There is a couple things to figure out that as well. I still need to work on a `Float32` routine and get that ironed out. We should only need the power series and asymptotic expansion. I think it might be confusing to readers if we have a primary routine that dispatch on the cutoff parameters but that some of the dispatches are never hit by the `Float32` routine as the first two branches cover all the cases.It is also the case right now that it was easier to design the cutoff for the Levin case before the power series case but that could be switched.

And I need to figure out what to do with the large negative values. This will still suffer from same issues on real line.

I don't really like throwing an error personally. I would prefer NaN (like SciPy) or extensively document the expected errors and return value with loss of precision (Boost Math Library). It seems more natural than to throw an error then return a value for -Inf. It also hurts the purity and ability to statically compile these functions with StaticCompiler (my primary interest with some of these routines)...

src/Airy/cairy.jl
```diff
@@ -263,29 +248,15 @@ function _airybiprime(z::Complex{T}) where T <: Union{Float32, Float64}
...
-        if iszero(y)
-            xabs = abs(x)
-            xx = T(2/3) * xabs * sqrt(xabs)
-            Jv, Yv = besseljy_positive_args(T(2)/3, xx)
-            Jmv = -(Jv + sqrt(T(3))*Yv) / 2
-            return convert(eltype(z), xabs * (Jv + Jmv) / sqrt(T(3)))
-        else
-            return -(cispi(T(2/3)) * _airybiprime(-z*cispi(one(T)/3)) + cispi(-T(2/3)) * _airybiprime(-z*cispi(-one(T)/3)))
-        end
+        return cispi(-5/6) * airyaiprime(z * cispi(-2/3)) + cispi(5/6) * airyaiprime(z*cispi(2/3))
```
oscardssmith, line 259: this will promote for `Float32`

src/Airy/cairy.jl
```diff
@@ -442,7 +402,63 @@ end
...
+            a = T(0.25) / (xsqr*x)
+
+            l = @ntuple $N i -> begin
+                out += t
+                t *= -3 * a * (i - 5//6) * (i - 1//6) / i
+                invt = @fastmath inv(t)
+                Vec{4, T}((reim(out * invt)..., reim(invt)...))
+            end
+            out = zero(typeof(x))
+            t = GAMMA_ONE_SIXTH(T) * GAMMA_FIVE_SIXTHS(T) / 4
```
oscardssmith, line 425: why the repetition?

src/Airy/cairy.jl
```diff
@@ -442,7 +402,63 @@ end
...
+            a = T(0.25) / (xsqr*x)
+
+            l = @ntuple $N i -> begin
+                out += t
+                t *= -3 * a * (i - 7//6) * (i + 1//6) / i
+                invt = @fastmath inv(t)
+                Vec{4, T}((reim(out * invt)..., reim(invt)...))
+            end
+            out = zero(typeof(x))
+            t = GAMMA_ONE_SIXTH(T) * GAMMA_FIVE_SIXTHS(T) / 4
```
oscardssmith, line 453: same here

src/Airy/cairy.jl
heltonmc: Ya will work on separate `Float32` implementations as they should never have to hit that branch.

src/Airy/cairy.jl
heltonmc: Just a product of getting it to be accurate right now. This should be able to be combined.

I'm thinking it will actually be great if we could split these functions into a separate function that can compute a sequence from some arbitrary series with a function handle. Therefore, we could avoid all these generated functions (just have two at the bottom that are repeatedly called). What's happening is we are generating all this code which is similar which is increasing compile times but in reality a lot of the instructions are overlapping it's just the series and `cumsum` that are different. The way it's laid out I think is poor. Hopefully, I'll come up with a better way to do that this week.

cgeoga:
Taking a look at the source of this PR as I play with the branch---just as a sanity check, do you mean for `besselkx_levin` to live in the `math.jl` file?

heltonmc:
> do you mean for `besselkx_levin` to live in the `math.jl` file?

No - just used that as an initial reference. I've moved it and added tests for those functions just to make sure they are working. I haven't made the optimizations but they are still reasonably fast right now.

Since this PR is focused on airy functions I'm going to finish those off and merge then I'll work more on the `besselk` case but you should be able to use these. I removed any premature breaking from them!

heltonmc:
Alright this finally completes the vision of having everything that doesn't depend on each other in separate modules. If someone in the future wants to split these off into separate packages that would be fine with me I think but for now this is good.

Supersedes #84. For example, now the AiryFunctions package just depends on the Math submodule which contains the math constants, levin transform, and reexports what we need from SIMDMath.

I'm sure there are some things to sort out which I will do depending on how we want to precompile some of the methods. Should precompile statements happen at the top module level or within each submodule that are then reexported? Hoping to take full advantage of v1.9 coming soon.

Still have several things to order out in the airy functions though now.

src/Airy/cairy.jl
heltonmc: I think I got it now with recent push. Only way I was able to get it to work right was to use @nexprs then combine with @ntuple later. This reduced the computational time by about 35% with recent pushes. Here is new update....
```julia
@generated function airybi_levin(x::Complex{T}, ::Val{N}) where {T <: Union{Float32, Float64}, N}
    :(
        begin

            xsqr = sqrt(x)
            xsqrx = xsqr * x

            t = GAMMA_ONE_SIXTH(T) * GAMMA_FIVE_SIXTHS(T) / 4
            t2 = 1 / t
            a = inv(4*xsqrx)
            a2 = 4 * xsqrx

            s = zero(typeof(x))
            s2 = zero(typeof(x))
            m = 1
            @nexprs $N i -> begin
                s += t
                s2 += t * m
                t *= -a * (3 * (i - 5//6) * (i - 1//6) / i)
                t2 *= -a2 * (i / (3 * (i - 5//6) * (i - 1//6)))
                l_{i} = Vec{4, T}((reim(s * t2)..., reim(t2)...))
                w_{i} = Vec{4, T}((reim(s2 * t2 * m)..., reim(t2 * m)...))
                m *= -1
            end

            l = @ntuple $N i -> l_{i}
            w = @ntuple $N i -> w_{i}

            e = exp(-2/3 * x * sqrt(x))
            return (e*im*levin_transform(l) + 2*levin_transform(w)/e) / (sqrt(T(π)^3) * sqrt(xsqr))
        end
    )
end
```

Might be a way to continue improve this..... but it's come along way already :)

src/Airy/cairy.jl
heltonmc: Fixed with recent push

heltonmc:
A little concerned that the testing times for airy function module has increased from 3s to 7s. Though locally this still takes around 3s. I'm wondering if through CI there is a high cost of compilation that isn't being cached appropriately. I ran a few benchmarks locally and can't reproduce so I think that is ok.

src/AiryFunctions/cairy.jl
```diff
@@ -307,18 +301,26 @@ airy_large_argument_cutoff(z::ComplexOrReal{Float32}) = abs(z) > 4
...
-    else
-        out = airybix_large_args(z)
-    end
-    return out .* exp(T(2/3) * z * sqrt(z))
+    out = airybix_large_args(z)
+    return isfinite(z) ? out .* exp(T(2/3) * z * sqrt(z)) : out
 end
 
 @inline function airyaix_large_args(z::Complex{T}) where T
+    if ~isfinite(z)
```
oscardssmith, line 315: why `~isfinite`?

heltonmc: I must have forgot about the `isinf` command :man_shrugging:

oscardssmith: no you probably do want to use isfinite since nan's exist. I just forgot that `~` does the same as `!`.

heltonmc: Ugh ya you are right I had it at top level which would check for NaN. I need to make an explicit NaN check at the top now because the else branch is where the NaNs will end up and will repeatedly call the airy. Need to just check first to avoid any NaNs going through the routines

heltonmc:
Alright I think this is ready to merge now to make way for https://github.com/JuliaMath/Bessels.jl/issues/96

PR #100, opened 2023-05-12
DenseVector → AbstractVector

gaspardbb:
Addresses https://github.com/JuliaMath/Bessels.jl/issues/99 

Changes the type `DenseVector{T}` to `AbstractVector{T}`.

### DIFF
--- a/src/BesselFunctions/besseli.jl
+++ b/src/BesselFunctions/besseli.jl
@@ -451,7 +451,7 @@ function _besseli(nu::Integer, x::T) where T <: Union{Float32, Float64}
 end
 
 """
-    Bessels.besseli!(out::DenseVector{T}, ν::AbstractRange, x::T)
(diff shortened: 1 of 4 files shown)


heltonmc:
Thanks for the very detailed and subsequent PR 😄 This is quite an interesting use case but I can see where this is advantageous (and even when it would be preferred). 

This looks good to me. The reason for this was I think when we were writing this that one blog post came out about one based indexing. My original implementation of this indexed into the allocated array at index `1` so an `AbstractVector` type would not technically be correct (or rather the implementation assumed the vector was one based index). I believe during the development of this feature we updated it to be more general but kept the DenseVector requirement. All that to say I think this is ok!

Thank you very much for the contribution 👍

heltonmc (approved):

heltonmc:
Ha I meant to say thank you for the very detailed issue and this PR!! I was just trying to keep my responses in a single thread but the issue was very helpful understanding your use case!

PR #101, opened 2023-05-22
An AD compatible besselk implementation for any order or argument (real values)

heltonmc:
This PR hopefully completes the `besselk` implementation to be fully AD compatible. There will of course be some things to continue to check with the derivative results but that should be fully on the AD side that we can solve those things. Perhaps the only thing will be to alter cutoffs if the derivative information is not as accurate.

Lot's of things to continue to go through. This implementation is actually more accurate and faster than the previous version for all cases except for arguments `x<1.5` and for integer orders. This is really the cost of having it be AD compatible to not rely on forward recurrence using `besselk0` and `besselk1` and have to rely on the more complicated series for these values. 

A good testing framework for this is tough.... Arb really struggles with `besselk` in general requiring a really high precision that crashes my computer each time for large arguments. The AMOS routines are pretty good so we could at least continue checking against those. Will have to work through that more.

Some benchmarks.
```julia
# this PR
julia> @benchmark Bessels.besselk(v, x) setup=(v=rand()*20; x=rand()*20)
 Range (min … max):  21.837 ns … 134.204 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     52.585 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   72.729 ns ±  24.429 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▁▁                █ ▂             ▂▁     ▁▁        ▂▅▆▄▄▄ ▂
  ▅▆▇███▇▅▃▁▁▁▁▁▁▁▁▁▁▁▁█▆█▇▇▇▇▆▅▇▆▇███▇████▆█████▅▁▁▁▁▃███████ █
  21.8 ns       Histogram: log(frequency) by time       106 ns <

# master

julia> @benchmark Bessels.besselk(v, x) setup=(v=rand()*20; x=rand()*20)
BenchmarkTools.Trial: 10000 samples with 982 evaluations.
 Range (min … max):   54.990 ns … 195.010 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     120.163 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   120.468 ns ±  18.256 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                               ▃▁▇▃▂█▆▇▅▅▅▇▅▂  ▁  ▃ ▃▃▁▂         
  ▂▂▂▂▁▂▃▂▂▁▃▄▂▂▄▃▂▃▅▃▃▄▆▃▆▄▅▆████████████████▇█▇████████▆▆▅▄▃▃ ▅
  55 ns            Histogram: frequency by time          157 ns <

```

### DIFF
--- a/src/BesselFunctions/besselk.jl
+++ b/src/BesselFunctions/besselk.jl
@@ -232,28 +232,107 @@ _besselk(nu::Union{Int16, Float16}, x::Union{Int16, Float16}) = Float16(_besselk
 
 _besselk(nu::AbstractRange, x::T) where T = besselk!(zeros(T, length(nu)), nu, x)
 
-function _besselk(nu::T, x::T) where T <: Union{Float32, Float64}
-    isinteger(nu) && return _besselk(Int(nu), x)
-    abs_nu = abs(nu)
-    abs_x = abs(x)
+function _besselk(v::T, x::T) where T <: Union{Float32, Float64}
+    v = abs(v)
+    x < 0 && return throw(DomainError(x, "Complex result returned for real arguments. Complex arguments are currently not supported"))
 
-    if x >= 0
-        return besselk_positive_args(abs_nu, abs_x)
+    if besselk_large_args_cutoff(v, x) return besselk_large_args(v, x)
+
+    elseif besselkx_levin_cutoff(v, x) return besselkx_levin(v, x, Val(16)) * exp(-x)
+
+    elseif besselik_debye_cutoff(v, x) return besselk_large_orders(v, x)
+    
     else
-        return throw(DomainError(x, "Complex result returned for real arguments. Complex arguments are currently not supported"))
-        #return cispi(-abs_nu)*besselk_positive_args(abs_nu, abs_x) - besseli_positive_args(abs_nu, abs_x) * im * π
+        v_floor, v_int = modf(v)
+        if x >= 1.5 # determine cutoff as function for differnet types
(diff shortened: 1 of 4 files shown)


src/BesselFunctions/besselk.jl
```diff
@@ -605,78 +504,110 @@ end
...
-    end
-    (kv, kvp1) = besselk_power_series_temme_basal(_v, x)
-    twodx = 2/x
-    for _ in 1:flv
-        _v += 1
-        (kv, kvp1) = (kvp1, muladd(twodx*_v, kvp1, kv))
-    end
-    kv
+@inline function f0_local_expansion_v0(v, x)
+    l2dx = log(2 / x)
```
oscardssmith, line 605: `log(2)-log(x)`?

heltonmc: Good point! We know for certain that `x < 1.5` for this routine so could check this. I want to set up some better testing and will change this

heltonmc:
Here is a plot of relative errors for the interesting region...

# master

Screen Shot 2023-05-22 at 2 15 33 PM

# this pr

Screen Shot 2023-05-22 at 2 15 50 PM

The errors along x axis is Arb giving bad results (increasing precision in Arb crashes my computer).

Idk what is going on with the errors in the enzyme stuff will have to track that down next.

PR #120, opened 2025-04-24
make float32 bessels type stable

oscardssmith:
These are not good implementations of these functions for Float32 (since they do all the math in Float64 to Float64 tolerances before converting back), but they are at least type stable (and as a bonus, now work with CUDA.jl)


heltonmc:
Ya... the Float32 implementations on some of these functions is definitely lacking. Would like to see some more pure implementations but these changes look good to me!

oscardssmith:
shall we merge as is and wait for further improvements?

oscardssmith:
Merging soonish unless there's objections

PR #128, opened 2026-04-06
Add Julia native loggamma functionality

TSGut:
In light of discussions at https://discourse.julialang.org/t/ann-meijerg-jl-a-julia-package-for-calculating-meijer-g-functions/136577 and locally in this repo at https://github.com/JuliaMath/Bessels.jl/issues/27, here is an adaptation of loggamma based on pre-existing as well as my recent work in SpecialFunctions.jl (https://github.com/JuliaMath/SpecialFunctions.jl/pull/522).

loggamma for reals and bigfloat calls external libraries in SpecialFunctions.jl, so I adapted Julia native versions as part of this. The real bigfloat version just reroutes to the Complex BigFloat one for now but if we care that can be changed for some significant speedups in that route in the future.

I don't know where the best place is to discuss where this goes - the issue, Julia discourse, etc. Or maybe this can restart conversations about a renaming of this package to something more general and thus SEO friendly?

Very happy for feedback or comments.


TSGut:
Alright, tests pass, coverage passes. The only thing failing is invalidations which as far as I can tell has nothing to do with my PR, it didn't add invalidations I think there is some version incompatibility in the action which causes an error?

@heltonmc  What's the PR process for this package? I would be happy for a review or suggestions.

Also @juliohm and @oscardssmith since they were part of the chat on Julia Discourse.

Things I intend to leave for the future for now:
1) Dedicated real BigFloat branch rather than using the slower Complex{BigFloat} logic.
2) Export loggammas and gammas? Rebrand? Make Gamma.jl elsewhere? Not sure which if any of this is desirable.

juliohm:
Thank you for leading this @TSGut ! @oscardssmith is more qualified to help with a review.

Hope this initiative will get the attention it deserves from package authors and maintainers in JuliaMath 🙏🏽

TSGut:
Benchmarking the implementations with the strategy suggested by the readme I get (compared to ArbNumerics.jl):

| Function        | Type        | Metric               | Value                      |
|-----------------|-------------|----------------------|----------------------------|
| loggamma        | Float64     | Mean relative error  | 2.602362769721367e-16      |
| loggamma        | Float64     | Max relative error   | 3.674838211509268e-14      |
| loggamma        | ComplexF64  | Mean relative error  | 1.3655212256834891e-16     |
| loggamma        | ComplexF64  | Max relative error   | 4.869646413689727e-16      |
| logabsgamma     | Float64     | Mean relative error  | 2.602362769721367e-16      |
| logabsgamma     | Float64     | Max relative error   | 3.674838211509268e-14      |

And execution time is roughly the same as using SpecialFunctions.jl:

| Function       | Type        | Library           | Mean (ns) | Median (ns) | Std (ns) | Speedup |
|----------------|-------------|--------------------|-----------|-------------|----------|---------|
| loggamma       | Float64     | Bessels            | 8.87      | 8.84        | 0.17     | 1.011×  |
| loggamma       | Float64     | SpecialFunctions   | 9.00      | 8.94        | 0.16     | 1.000×  |
| loggamma       | ComplexF64  | Bessels            | 28.79     | 28.78       | 0.14     | 1.010×  |
| loggamma       | ComplexF64  | SpecialFunctions   | 29.08     | 29.05       | 0.22     | 1.000×  |
| logabsgamma    | Float64     | Bessels            | 10.49     | 10.54       | 0.23     | 0.832×  |
| logabsgamma    | Float64     | SpecialFunctions   | 8.80      | 8.77        | 0.30     | 1.000×  |

src/GammaFunctions/loggamma.jl
```diff
@@ -0,0 +1,351 @@
...
+loggamma(z::Complex{<:Integer}) = _loggamma(Complex{Float64}(z))
+loggamma(z::Complex{<:Rational}) = loggamma(float(z))
+function loggamma(x::BigFloat)
+    if isnan(x)
+        return x
+    elseif isinf(x)
+        return x > 0 ? x : BigFloat(NaN)
+    elseif x <= 0
+        x == 0 && return BigFloat(Inf)
+        sinpi(x) == 0 && return BigFloat(Inf)  # negative integer pole
```
oscardssmith:
```suggestion
        isinteger(x) && return BigFloat(Inf)  # negative integer pole
```

src/GammaFunctions/loggamma.jl
```diff
@@ -0,0 +1,351 @@
...
+        return x > 0 ? Inf : NaN
+    elseif x ≤ 0
+        x == 0 && return Inf
+        s = sinpi(x)
+        s == 0 && return Inf  # negative integer pole
+        # reflection: log|Γ(x)| = log(π) - log|sin(πx)| - log(Γ(1-x))
+        # but loggamma for real requires Γ(x)>0
+        y, sgn = _logabsgamma(x)
+        sgn < 0 && throw(DomainError(x, "`gamma(x)` must be non-negative"))
+        return y
```
oscardssmith, line 165: we likely want a version of `_loggamma` that skips the duplicate checks that `logabsgamma` performs

src/GammaFunctions/loggamma.jl
```diff
@@ -0,0 +1,351 @@
...
+    elseif x < 7
```
oscardssmith, line 166: is there a polynomial/rational that works for this region?

src/GammaFunctions/loggamma.jl
```diff
@@ -0,0 +1,351 @@
...
+        end
+    elseif x > 7 || yabs > 7
+        return _loggamma_asymptotic(z)
+    elseif x < 0.1
+        if x == 0 && y == 0
+            return Complex(Inf, signbit(x) ? copysign(Float64(π), -y) : -y)
+        end
+        # reflection formula with correct branch cut.
+        return Complex(LOGPI_F64, copysign(TWO_PI_F64, y) * floor(0.5 * x + 0.25)) -
+            log(sinpi(z)) - _loggamma(1 - z)
```
oscardssmith, line 219: is `log(sinpi(z))` stable?

TSGut: All my tests on it suggest that it is. Do you have a specific stress test in mind? It behaves very well even near poles and is what SpecialFunctions.jl uses as well. If you mean choice of branchcut wise, that's what that line ensures.

oscardssmith: the worry would be places where sinpi(z)~1 since then `log(sinpi(z))` ~0. This will only be a problem if there is a place where `sinpi(z)~1` and `loggamma(1-z)~0`, and it would manifest as poor relative error near a zero of the function.

TSGut: Let's see...

```julia
using Bessels
using ArbNumerics
using Printf

function arb_loggamma(z)
    za = ArbComplex(real(z), imag(z))
    return log(ArbNumerics.gamma(za))
end

# Points where sin(pi*z) = 1
zs = [1/2 + 2k for k in -10:10]

firstrun = true
for z in zs
    if firstrun
        @printf("%8s  %12s  %20s  %18s\n",
                "z", "sin(pi*z)", "loggamma(1-z)", "rel_error")
        @printf("%8s  %12s  %20s  %18s\n",
                "--------", "------------", "--------------------", "------------------")
        firstrun = false
    end

    s   = sinpi(z)
    lg1 = loggamma(1 - z)
    lg  = loggamma(z)
    lgr = arb_loggamma(z)

    denom = abs(lgr) > eps() ? abs(lgr) : eps()
    relerr = abs(lg - lgr) / denom

    @printf("%8.2f  %12.6f  %20.12e  %18.3e\n",
            z, s, lg1, relerr)
end
```

gives

| z      | sin(πz) | loggamma(1−z)          | rel_error |
|--------|---------|-------------------------|-----------|
| -19.50 | 1.000000 | 4.083150097453e+01     | 3.137e-17 |
| -17.50 | 1.000000 | 3.494331577688e+01     | 9.067e-17 |
| -15.50 | 1.000000 | 2.927775451504e+01     | 1.281e-16 |
| -13.50 | 1.000000 | 2.386276584169e+01     | 1.876e-17 |
| -11.50 | 1.000000 | 1.873434751194e+01     | 1.177e-16 |
| -9.50  | 1.000000 | 1.394062521940e+01     | 6.760e-17 |
| -7.50  | 1.000000 | 9.549267257301e+00     | 1.743e-16 |
| -5.50  | 1.000000 | 5.662562059857e+00     | 8.885e-17 |
| -3.50  | 1.000000 | 2.453736570842e+00     | 9.157e-16 |
| -1.50  | 1.000000 | 2.846828704729e-01     | 1.513e-15 |
| 0.50   | 1.000000 | 5.723649429247e-01     | 2.531e-15 |
| 2.50   | 1.000000 | 8.600470153765e-01     | 4.606e-15 |
| 4.50   | 1.000000 | -1.309006684993e+00    | 4.927e-16 |
| 6.50   | 1.000000 | -4.517832174008e+00    | 1.119e-16 |
| 8.50   | 1.000000 | -8.404537371452e+00    | 8.471e-17 |
| 10.50  | 1.000000 | -1.279589533355e+01    | 1.500e-17 |
| 12.50  | 1.000000 | -1.758961762609e+01    | 1.455e-16 |
| 14.50  | 1.000000 | -2.271803595584e+01    | 4.535e-17 |
| 16.50  | 1.000000 | -2.813302462919e+01    | 1.007e-16 |
| 18.50  | 1.000000 | -3.379858589103e+01    | 6.893e-17 |
| 20.50  | 1.000000 | -3.968677108868e+01    | 1.442e-17 |

So I think it's fine?

oscardssmith: Isn't this the complex case? Isn't there a whole surface along which `sinpi` is zero?

oscardssmith: Edit: right. `sin` only has real zeros

TSGut: I thought so but you did make me double check there so it was a good call to sanity check it! 😄

src/GammaFunctions/loggamma.jl
TSGut: I added specialised unsafe versions of `loggamma` and `logabsgamma` for Float64 tailored to the scenarios where they call each other to reduce duplicate checks. Let me know if you think that doesn't resolve this.

src/GammaFunctions/loggamma.jl
oscardssmith: I guess the code should probably have a comment so future readers know we thought about it.

src/GammaFunctions/loggamma.jl
TSGut: Hm, well the region from 0 to 1 definitely needs this, see https://www.wolframalpha.com/input?i=loggamma%28x%29+from+0+to+7

We don't want to approximate that left side. 

But the region from 1 to 7 one could split into two or three intervals and do hard coded polynomial evaluation on. The 'issue' is that this needs maybe 15-20 coefficients to get same accuracy on dense sweep on each of the intervals. It works, though. Performance wise it's irrelevant because we trade a smaller polynomial evaluation and less region checks + `log(prod)` for higher degree polynomial eval and more region checks.

Any preference?

oscardssmith: A rational approximation should be able to get the zero to 1 region I think? either way, this is fine for now.

TSGut:
@oscardssmith: I think this covers everything in your code review -- anything else left to do before this can be merged?

I leave it to you whether you choose to export these functions but the sooner this can be tagged the sooner I can get started on making the hypergeometric functions be Julia native. 😄 I have had a chat with the maintainer of HypergeometricFunctions.jl and he's happy for me to basically copy and adjust his code as functions within MeijerG.jl, so MeijerG.jl will just contain its own native hypergeometric function logic and probably also export a Julia native pFq, with its only dependency becoming Bessels.jl.

oscardssmith:
The code looks good to me, but I do think or should probably live in a new Gamma.jl. there are a ton more people that likely will want a good gamma function (e.g. Primes.jl) that probably don't want to pull in all the bessel functions as well.

TSGut:
I am happy to do that, can you make a Gamma.jl in JuliaMath (we should check that no such named package is currently registered)? While we're at it could I be added to JuliaMath - would make this easier going forward.

I'll move this code into a PR for the new Gamma.jl then.

Also, if this will create a new Julia native Special functions ecosystem then we probably want to have references pointing at the other relevant packages in that ecosystem from the readmes or docs.

oscardssmith:
https://github.com/JuliaMath/Gamma.jl. If you make the PR, I'll work on getting this to be an actual Julia repo.

TSGut:
Great, I will close this PR since we are moving development there.

PR #129, opened 2026-04-11
remove Gamma from Bessels.jl and add dep to Gamma.jl

oscardssmith:
Now that Gamma.jl is registered, we don't need to have our own Gamma impl. As a plus, Gamma.jl also has `loggamma` so we could switch some of the algorithms that we avoided do the the lack of `loggamma`.


oscardssmith:
@heltonmc do you have thoughts here? If not, I'll likely merge in a couple days.

